PR #37, opened 2022-11-16
Make settings hashable

dwhswenson:
Looks like they're not hashable by default? I'm getting complaints from `lru_cache`

### DIFF
--- a/fah_alchemy/settings.py
+++ b/fah_alchemy/settings.py
@@ -19,8 +19,8 @@ class Settings(BaseSettings):
     JWT_EXPIRE_SECONDS: int = 1800
     JWT_ALGORITHM: str = 'HS256'
 
-    def __hash__(self):
-        return hash(tuple(self.dict().items()))
+    class Config:
+        frozen = True
 
 
 @lru_cache()
--- a/fah_alchemy/settings.py
+++ b/fah_alchemy/settings.py
@@ -19,6 +19,9 @@ class Settings(BaseSettings):
     JWT_EXPIRE_SECONDS: int = 1800
     JWT_ALGORITHM: str = 'HS256'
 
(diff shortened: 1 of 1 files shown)


dotsdl:
Wait, where are we seeing complaints? I haven't seen any in my test outputs; can you show me?

fah_alchemy/settings.py
```diff
@@ -19,6 +19,9 @@ class Settings(BaseSettings):
     JWT_EXPIRE_SECONDS: int = 1800
     JWT_ALGORITHM: str = 'HS256'
 
+    def __hash__(self):
+        return hash(tuple(self.dict().items()))
+
```
dotsdl, line 24: Does changing this to:

```
    class Config:
        frozen = True
```

fix the warnings you're seeing?

dwhswenson:
MCVE (without this fix):

```python
import os
from fah_alchemy.settings import get_settings
from fah_alchemy.compute.api import get_n4js

args = {"NEO4J_URL": "https://foo", "NEO4J_USER": "me", "NEO4J_PASS": "password", "JWT_SECRET_KEY": "secret"}
os.environ.update(args)
settings = get_settings()
get_n4js(settings)
```

```pytb
TypeError                                 Traceback (most recent call last)
Cell In [1], line 8
      6 os.environ.update(args)
      7 settings = get_settings()
----> 8 get_n4js(settings)

TypeError: unhashable type: 'Settings'
```

This is because `get_n4js` uses `@lru_cache`, with the key being the unhashable settings object.

Using `class Config: frozen=True` also fixes; I'll switch this PR to using that. (Obviously, the MCVE still errors, since there's no DB configured for those settings, but that is the error I expect).

PR #38, opened 2022-11-21
CLI: Launch APIs

dwhswenson:
- [x] `fah-alchemy api`
- [x] `fah-alchemy compute api`
- [x] tests for both

Seems to work when I run it locally, but I haven't quite got the testing figured out yet

@dotsdl : It would be useful if you could tell me which settings (if any) are relevant for each API launch. Then I'll add those as options.


dotsdl:
Perhaps add one option to allow the JWT SECRET KEY to be generated on startup. This would just be a convenience for the single API service instance case; it would not be appropriate if launching many separate gunicorn instances.

dotsdl:
Planning to hit this today @dwhswenson; anything additional I should know?

dwhswenson:
There's probably some cleanup to be done here about getting the host URL / port from env vars as well; I can probably help with those parts, but we have a few things I'm focusing on from the OpenFE side for the next week (until board meeting).

Main issue is that I think your existing fixtures must mock out something that can't be pickled, and I couldn't trace down what that was. Without that, multiprocessing choking on the subprocess call.

dotsdl:
@dwhswenson I believe the APIs now start up just fine with the tests. I noticed that they weren't dying properly, so the `test_api_application` test's service was sticking around by the time `test_api` would run. It doesn't look like this issue persists after my latest changes.

dotsdl:
Would you mind wiring up env variables into the API CLI points? I've added the appropriate `*Settings` import into each point to indicate which one applies, and some pointers for how to inject these into the app in the user-facing API CLI entry point.

dotsdl:
Thanks for doing this @dwhswenson! What you had was basically working already; very slick!

dotsdl:
Looks like CI is just generally failing now, not just for this PR: https://github.com/openforcefield/fah-alchemy/actions/runs/3590301180

Unclear if this is due to changes on GHA that we'll have to find workarounds for, or if this is just some temporary condition that will resolve itself. Tests pass locally just fine in any case.

dotsdl:
Think I fixed the tests; was due to the changes in `gufe` on tokenization. I've included my fixes here: https://github.com/OpenFreeEnergy/gufe/pull/102

dotsdl:
Thanks @dwhswenson! This looks great; merging!

PR #46, opened 2022-11-29
[WIP] Adding tutorial for building AlchemicalNetworks for consumption with `fah-alchemy`

dotsdl:
This PR adds the sphinx doc skeleton, along with our first tutorial. The
scope of this tutorial should be AlchemicalNetwork creation, up to
submission using the `FahAlchemyClient`.


dotsdl:
Would like to use a real `Protocol` for this, which includes creating `Settings`. Need to establish pattern for users for creating `Settings`.

JenkeScheen:
I did a run-through of this draft, some quick comments. They are mostly `gufe`-related, the `AlchemicalNetwork` operations LGTM!

1. From a `ProteinComponent`, what is the recommended way of returning the number of atoms in the molecule? This is super helpful for debugging in these kinds of workflows. `len(list(protein.to_openmm_positions()))` seems convoluted.
2. For `Settings(.., forcefield_file="foobar.xml", ..)`, will this file live on `lilac` or `F@H`? Would it be worth absorbing some default ones into `fah-alchemy`?
3. Are there plans to implement some sort of atom-mapping visualiser and make it requestable from e.g. a `Transformation()`? This is currently implemented in `Perses`: https://github.com/choderalab/perses/blob/main/examples/atom-mapping/Atom-mapping.ipynb
4. Between `complexed_transformations` and `solvated_transformations` there should be some check in place that makes sure all atom-mappings are the same for all edges between the two phases, or at least give a warning if they are not. 
5. Finally a more infrastructural Q: at least when we're deploying this on `lilac`, where would this notebook be run? Would it work just fine on an interactive node or is the intention to run locally, then send it off to `lilac`?

IAlibay:
> Are there plans to implement some sort of atom-mapping visualiser

Just a point on this, these are current available under openfe (including an upcoming py3dmol based one), and I think we came to the conclusion that we didn't think pulling the necessary dependencies should be in gufe. I'm not sure if moving them makes sense, probably worth discussing.

dotsdl:
Per discussion with @IAlibay, may make sense to migrate this tutorial to `gufe`.

dotsdl:
@richardjgowers will take over this PR, migrate to `gufe`.

hmacdope:
Would people be okay if I split the sphinx docs setup for FAH off this PR? I might do that now so I can work on some of the docs?

dotsdl:
@hmacdope yes please! Go for it!

dotsdl:
I'll create a PR into `gufe` with this notebook, link to the discussion here so @JenkeScheen's comments aren't lost.

dotsdl:
Will instead make a PR into openfreeenergy/ExampleNotebooks.

dotsdl:
Superseded by https://github.com/OpenFreeEnergy/ExampleNotebooks/pull/36

PR #53, opened 2022-12-12
Implement `Scope` enforcement in API

### LINKED ISSUE #41: Add `Scope` enforcement by APIs for user and compute identities
[User and compute identities](https://github.com/openforcefield/fah-alchemy/blob/0cdfa891cddbc55cf8d6daaaded719903dc89ad4/fah_alchemy/security/models.py#L22-L41) should feature a list of [`Scope`s in `str`](https://github.com/openforcefield/fah-alchemy/blob/main/fah_alchemy/models.py#L6-L16) form, which go like `<org>-<campaign>-<project>`. 

For user identities, the `FahAlchemyAPI` should deny access to objects in `Scope`s that are not represented in that user's list of allowed `Scope`s. Likewise, for compute identities, the `FahComputeAPI` should do the same.

This provides basic boundaries for multi-tenant use of a single deployment. It is critical if we wish to proceed with a single deployed system for OpenFF and ASAP.

Lnaden:
This mostly resolves #41.

Checks against the `Scope` of the `Token` submitted to the API are done to ensure no one can access networks which are not provided through their Token. Many other changes have expanded testing framework to make future iterations and features involving `Scope` checking quicker to implement and test.

Main caveat/concern is on testing a Token Data identity is just created for our test user and I don't know how that translates into end-users just being able to re-create the correct token data if need be? Someone who knows the FastAPI and the Neo4j system better can verify for me if this is the level of security desired.

The test suite has been improved to include more general fixtures and more general prepared Neo4j system to test against multiple networks and scopes.

The API has been updated to check against incoming scope validation through a regular scheme. `query_networks` has a rudamentary implementation, but I have not fully tested it since the client-side API method has not been impemented yet.

Other miscelaneous changes:
* Added stringification methods to the `Scope` model to be inline with the `ScopedKey`.
* Added `return_gufe` to the `query_network` method of `statestore` to enable API logic to work when implemented as is.
* Fixed testing on OSX platform due to multiprocessing handling of nested methods with the `set_start_method` to duplicate `Unix`'s default. This might overlap with changes in another PR but big thanks to @dotsdl and @dwhswenson for consultation and debugging help.
* Created a more consistent set of scopes to test against across test framework setup
* Added tests for most new feature implemented and their fail states (exception `query_networks` for reason above).

Ready for review and comments

### DIFF
--- a/devtools/conda-envs/test.yml
+++ b/devtools/conda-envs/test.yml
@@ -30,6 +30,7 @@ dependencies:
   ## api(s)
   - fastapi
   - uvicorn
+  - gunicorn
   - python-jose
   - passlib
   - bcrypt
--- a/fah_alchemy/base/api.py
+++ b/fah_alchemy/base/api.py
@@ -72,7 +72,9 @@ def validate_scopes_query(
         acc_scope = Scope.from_str(accessible_scope)
         add_it_in = True
 
-        for (query_field, target_field) in zip(query_scope.to_tuple(), acc_scope.to_tuple()):
+        for (query_field, target_field) in zip(
+            query_scope.to_tuple(), acc_scope.to_tuple()
+        ):
 
             # match (query_org == token_org) then (query_campaign == token_campaign) then (query_proj == token_proj)
             if not (query_field == target_field or query_field is None):
@@ -83,9 +85,7 @@ def validate_scopes_query(
                 break
 
         if add_it_in:
-            scope_intersection.append(
-                acc_scope if not as_str else accessible_scope
-            )
+            scope_intersection.append(acc_scope if not as_str else accessible_scope)
 
     return scope_intersection
 
@@ -93,7 +93,7 @@ def validate_scopes_query(
 class QueryGUFEHandler:
     """
     Helper class to provide a single-dispatch like handling of the query
-    operations since they can return list or dict. 
+    operations since they can return list or dict.
     """
 
     def __init__(self, return_gufe: bool):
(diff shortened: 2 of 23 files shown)


dotsdl:
Thanks @Lnaden! Reviewing now!

dotsdl:
Looking great @Lnaden! I made some adjustments in 298afb0, building on your work here. Really welcome additions to the package.

Would you be willing to:
- add use of `validate_scopes` to the compute API endpoints as well, with tests?
- pull the contents into its own function, perhaps `validate_scopes_query`, that would be used for other `query_*` API endpoints?

I can take these on if you're short on time, but would very much appreciate your effort on these if at all possible!

Lnaden:
I can. I'll have some time this week and should get it sometime Friday afternoon at the latest. I'll let you know if something changes.

dotsdl:
Awesome, thank you so much! Please let me know if any clarifications are needed!

Lnaden:
Expand testing and scope checks to meet request

* Update general API internal utility functions to generalize query_* functions. Also made a GUFE handler class to make the query codes more single-dispatch-like 
* Add validation against scope for Compute API calls, best I could given a number of non-implemented methods.
* Upgraded Compute test framework to align with the Interface test network. This includes multiple scopes and nodes in the Neo4j graph. 
* Updated the client to authenticate against a single scope to allow checking against scope and fail-state scopes.
* Renamed the general test users to generally indicated what part of the code they are testing against.

I think this meets your requested changes, @dotsdl. There's lots more tests which could be written, but its not exactly clear how to best call some of the code without sitting down and thinking about it more. However, all the framework is there to quickly implement new tests. Let me know what you think.

dotsdl:
Awesome, thanks @Lnaden! Planning to hit this today!

dotsdl (approved):
Looks great @Lnaden! I added a couple unit tests for the new functions, and otherwise trimmed down `validate_scopes_query` since we can be sure we already get a functioning `Scope` object for `query_scope`. Also added validation for `Scope` fields themselves in the model.

Really substantial effort here; thank you again for it!

PR #59, opened 2023-01-12
Add API connectivity check method to `FahAlchemyClientBase`

### LINKED ISSUE #55: Add an underscore method to `FahAlchemyClient` and `FahAlchemyComputeClient` that checks connectivity of API to storage components
We would like a `_api_check` method added to the `FahAlchemyBaseClient` that hits a `/check` endpoint on its target API. The `/check` endpoint on each of `FahAlchemyAPI` and `FahAlchemyComputeAPI` should be implemented to hit both the state store (Neo4j) and the object store (S3) in such a way that it simply checks that these services are meaningfully reachable. 

The `/check` endpoint should return a `200 OK` HTTP status if the services are reachable, and an appropriate status code if they are not. The `_api_check` method on each client should raise an exception if any status code other than `200 OK` comes back.

These methods would be used for checking that a deployment via e.g. `docker-compose` is up and working as expected.

hmacdope:
Fixes #55

Adds a method on `FahAlchemyClientBase` to poll endpoints on `FahAlchemyAPI` and `FahAlchemyComputeAPI` that check the status of the `ObjectStore` and `StateStore`

Notes:

I returned the status code as part of the `JSON` payload, I know this is probably not the right way to do it but wasn't exactly sure what the best way was to return both the payload and the status code.

### DIFF
--- a/fah_alchemy/base/api.py
+++ b/fah_alchemy/base/api.py
@@ -160,7 +160,8 @@ def _check_store_connectivity(n4js: Neo4jStore, s3os: S3ObjectStore) -> dict:
         raise HTTPException(
             status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=detail
         )
-    return status.HTTP_200_OK
+    else:
+        return status.HTTP_200_OK
 
 
 async def get_token_data_depends(
--- a/fah_alchemy/base/api.py
+++ b/fah_alchemy/base/api.py
@@ -147,8 +147,9 @@ def scope_params(org: str = None, campaign: str = None, project: str = None):
             headers={"WWW-Authenticate": "Bearer"},
         )
 
+
 def _check_store_connectivity(n4js: Neo4jStore, s3os: S3ObjectStore) -> dict:
-    """ Check if neo4j and s3 object store are reachable """
+    """Check if neo4j and s3 object store are reachable"""
     # check if neo4j database is reachable
     neo4jreachable = n4js._store_check()
     # check if s3 object store is reachable
@@ -156,8 +157,11 @@ def _check_store_connectivity(n4js: Neo4jStore, s3os: S3ObjectStore) -> dict:
 
     if not neo4jreachable or not s3reachable:
         detail = f"Attempt to reach services failed, Neo4j reachable: {neo4jreachable}, S3 reachable: {s3reachable}"
-        raise HTTPException(status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=detail)
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=detail
+        )
     return status.HTTP_200_OK
+
 
 async def get_token_data_depends(
     token: str = Depends(oauth2_scheme),
--- a/fah_alchemy/base/api.py
+++ b/fah_alchemy/base/api.py
@@ -147,6 +147,21 @@ def scope_params(org: str = None, campaign: str = None, project: str = None):
             headers={"WWW-Authenticate": "Bearer"},
         )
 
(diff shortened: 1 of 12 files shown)


fah_alchemy/compute/api.py
```diff
@@ -63,6 +63,22 @@ async def info():
...
+):
+    # check if neo4j database is reachable
+    neo4jreachable = n4js._store_check()
+    # check if s3 object store is reachable
+    s3reachable = s3os._store_check()
+    if not neo4jreachable or not s3reachable:
+        code = status.HTTP_503_SERVICE_UNAVAILABLE
+    else:
+        code = status.HTTP_200_OK
+    return {"neo4jreachable": neo4jreachable, "s3reachable": s3reachable, "code": code}
```
hmacdope, line 79: @dotsdl This is returning a status code as part of a JSON payload, I know this is probably not the right way to do it but wasn't exactly sure what the best way was to return a payload and status code so that `requests` picks it up.

fah_alchemy/interface/api.py
```diff
@@ -60,6 +60,22 @@ async def info():
...
+):
+    # check if neo4j database is reachable
+    neo4jreachable = n4js._store_check()
+    # check if s3 object store is reachable
+    s3reachable = s3os._store_check()
+    if not neo4jreachable or not s3reachable:
+        code = status.HTTP_503_SERVICE_UNAVAILABLE
+    else:
+        code = status.HTTP_200_OK
+    return {"neo4jreachable": neo4jreachable, "s3reachable": s3reachable, "code": code}
```
hmacdope, line 76: Same here for payload vs status code.

hmacdope (commented):
@dotsdl see my comment :)

fah_alchemy/compute/api.py
hmacdope: To elaborate I mean so that it is then accessible with `response.status_code.`

fah_alchemy/tests/integration/interface/client/test_client.py
```diff
@@ -51,3 +51,11 @@ def test_create_network(
         uvicorn_server,
     ):
         ...
+
+    def test_api_check(
```
dotsdl, line 55: Can you add this to `fah_alchemy/tests/integration/compute/client/test_compute_client.py` as well?

dotsdl (changes requested):
Thanks for this @hmacdope! Looking really good, but I made some suggestions for changes. Can you give these a go before we proceed to merge?

hmacdope:
@dotsdl  I think I have addressed your comments :)

fah_alchemy/base/api.py
```diff
@@ -148,6 +148,22 @@ def scope_params(org: str = None, campaign: str = None, project: str = None):
...
+    # check if s3 object store is reachable
+    s3reachable = s3os._store_check()
+
+    if not neo4jreachable or not s3reachable:
+        detail = f"Attempt to reach services failed, Neo4j reachable: {neo4jreachable}, S3 reachable: {s3reachable}"
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail=detail
+        )
+    else:
+        return status.HTTP_200_OK
```
dotsdl: We don't need to return a status code, and this isn't how `fastapi` handles them. I'll return `True` here instead just to indicate everything's good if no exception is raised.

dotsdl: The fastapi docs detail how to make an API endpoint give a different status code on success: https://fastapi.tiangolo.com/tutorial/response-status-code/ ; otherwise for failure the exception we raise here will apply.

dotsdl (approved):
Much improved, thank you @hmacdope! This is excellent work!

Merge when convenient!

hmacdope:
@dotsdl do we merge or squash merge?

PR #64, opened 2023-01-13
Add docs build and deploy workflow

### LINKED ISSUE #62: Documentation build workflow for CI
**In broad terms, what are you trying to do?**
We should have a github actions workflow that checks that the docs build. 

**Describe the new feature(s) that would help you to achive this**
A simple docs workflow should be added to check that the sphinx build works.

hmacdope:
Fixes #62 and adds docs deployment to Github Pages. This is intended as a first swipe at publishing the docs rather than a polished final product. 

 I chose github pages simply because its the easiest for me to set up having worked with it before, but as suggested by @mikemhenry, there is no vendor lock so if people would prefer RTD I am happy to use that instead.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,48 @@
+name: "Docs build and deploy"
+on: [push]
+
+defaults:
+  run:
+    shell: bash -l {0}
+
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
+    env:
+      OE_LICENSE: ${{ github.workspace }}/oe_license.txt
+
+    steps:
+      - name: checkout
+      - uses: actions/checkout@v3
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+            auto-update-conda: true
+            use-mamba: true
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,55 @@
...
+          cd docs
+          make html
+
+      - name: Deploy 🚀
+        if: github.event_name != 'pull_request'
+        uses: JamesIves/github-pages-deploy-action@3.7.1
+        with:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          BRANCH: gh-pages # The branch the action should deploy to.
+          FOLDER: docs/_build/html # The folder the action should deploy.
```
dotsdl, line 55: I take it the `if` statement will only actually deploy on merge into `main`? If so, that is the desired outcome I think.

dotsdl (approved):
Looks great @hmacdope! Feel free to merge if satisfied!

.github/workflows/docs.yml
hmacdope: Yeah any push to main should trigger deploy.

PR #66, opened 2023-01-17
Add CLI point for adding / removing compute identities

### LINKED ISSUE #63: Add CLI point for adding user / compute identities
We need to be able to add user and compute identities via the command-line interface on the hosting resource. This will require credentials for the state store, and should take as input the identity type (user, compute), identifier, and key.

hmacdope:
Fixes #63. 

- Adds a utility to add and remove user identities on the command line. 
- Adds some parameterisation to the tests of adding and removing credentialed identities.


hmacdope:
This should be ready for review now.

fah_alchemy/tests/integration/test_cli.py
```diff
@@ -250,3 +255,121 @@ def test_database_reset(n4js_fresh, network_tyk2, scope_test):
...
+    user_type_str, user_type_cls = user_type
+    env_vars = {
+        "NEO4J_URL": n4js.graph.service.uri,
+        "NEO4J_USER": "neo4j",
+        "NEO4J_PASS": "password",
+    }
+    runner = CliRunner()
+    with set_env_vars(env_vars):
+        ident = "bill"
+        key = "and ted"
```
dotsdl, line 275: :grin:

fah_alchemy/tests/integration/test_cli.py
```diff
@@ -250,3 +255,121 @@ def test_database_reset(n4js_fresh, network_tyk2, scope_test):
...
+
+        with pytest.raises(KeyError, match="No such object in database"):
+            cred = n4js.get_credentialed_entity(ident, user_type_cls)
+
+
+@pytest.mark.parametrize(
+    "user_type",
+    [("user", CredentialedUserIdentity), ("compute", CredentialedComputeIdentity)],
+)
+def test_user_remove_bad_cred(n4js_fresh, user_type):
```
dotsdl: We can remove this test since we shouldn't need an identity's key to delete it.

fah_alchemy/cli.py
```diff
@@ -364,30 +372,115 @@ def reset(url, user, password, dbname):
...
+    cli_values = url | user | password | dbname
+
+    settings = get_settings_from_options(cli_values, Neo4jStoreSettings)
+    n4js = get_n4js(settings)
+
+    identity_type_cls = _identity_type_string_to_cls(identity_type)
+    if authenticate(n4js, identity_type_cls, identifier, key):
+        n4js.remove_credentialed_identity(identifier, identity_type_cls)
+    else:
+        raise AuthenticationError("Authentication with StateStore failed.")
```
dotsdl, line 468: An admin that needs to remove an identity from a deployed `fah-alchemy` instance generally won't have the credential for that identity, e.g. a user's key. We should remove `key` as an argument from this CLI command.

fah_alchemy/tests/integration/test_cli.py
```diff
@@ -250,3 +255,155 @@ def test_database_reset(n4js_fresh, network_tyk2, scope_test):
...
+        assert click_success(result)
+        for ident in identities:
+            assert ident in result.output
+
+
+@pytest.mark.parametrize(
+    "identity_type",
+    [("user", CredentialedUserIdentity), ("compute", CredentialedComputeIdentity)],
+)
+def test_identity_remove_bad_cred(n4js_fresh, identity_type):
```
dotsdl, line 376: We can remove this test, since we don't want to require a credential be given for an identity we wish to remove.

dotsdl (approved):
Looking great @hmacdope! 

I changed `user` to `identity` for the CLI group these commands fall under, since we use the same CLI points for managing both user and compute identities (and in the future, admin identities). 

I've flagged one change that we need before we should merge. Once that's in I think this is good to go.

fah_alchemy/cli.py
hmacdope: Roger that, this was due to my misconception as to who would be using the CLI and where. 👍 Makes perfect sense.

dotsdl: No problem!

fah_alchemy/tests/integration/test_cli.py
hmacdope: Done

hmacdope:
All issues addressed so I think I will go ahead and merge.

PR #68, opened 2023-01-19
Add hierarchy to `Scope`

### LINKED ISSUE #42: Add Scope rules for hierarchy of org > campaign > project
Currently, `Scope` is just a triple of strings representing `org`, `campaign`, `project`, with no sense of hierarchy between these levels. This results in cumbersome access permissions for e.g. a user that should have access to an `org` called `openff`: they will require an explicit entry for each `campaign` and `project` represented in that `org` to have full access, with new entries added as new `campaign`s and `project`s are created.

To avoid this, we'd like for the hierarchy between these levels to be made explicit, and for glob-based wildcards to be allowed. So, we could have e.g. `*-*-*` (all scopes), `openff-*-*` (all `org` 'openff' scopes), `openff-rosemary-*` (all `org` 'openff', `campaign` 'rosemary' scopes), as allowed entries in a user's list of `Scope`s.

This hierarchy enforcement should disallow entries like `*-rosemary-*` (access to any `campaign` called 'rosemary', anywhere). In other words, one cannot limit scope at a level lower when given broadest scope at a level higher.

Ideally, `Scope` validity is enforced in the [Pydantic model itself](https://github.com/openforcefield/fah-alchemy/blob/0cdfa891cddbc55cf8d6daaaded719903dc89ad4/fah_alchemy/models.py#L6-L16) on creation.

hmacdope:
Partially fixes #42 

I have also added the `no-dash` validation for `ScopedKey`, as mentioned briefly in #42. Let me know if this is wrong and I can take it out. 

I also added a test for invalid scope creation with a dash in the name

I have also added support for "*" wildcards in addition to `None`

### DIFF
--- a/fah_alchemy/base/api.py
+++ b/fah_alchemy/base/api.py
@@ -36,7 +36,8 @@ from ..security.models import Token, TokenData, CredentialedEntity
 
 
 def validate_scopes(scope: Scope, token: TokenData) -> None:
-    """Verify that token data has specified Scope encoded."""
+    """Verify that token data has specified Scope encoded directly or is accessible via
+    scope hierarchy."""
 
     if not isinstance(scope, Scope):
         raise ValueError("`scope` must be a `Scope` object to ensure validity")
@@ -47,7 +48,7 @@ def validate_scopes(scope: Scope, token: TokenData) -> None:
         raise HTTPException(
             status_code=status.HTTP_401_UNAUTHORIZED,
             detail=(
-                f"Targeted scope '{scope}' not in found among scopes for this identity: {token.scopes}."
+                f"Targeted scope '{scope}' not accessible via scopes for this identity: {token.scopes}."
             ),
             headers={"WWW-Authenticate": "Bearer"},
         )
--- a/fah_alchemy/base/api.py
+++ b/fah_alchemy/base/api.py
@@ -66,10 +66,24 @@ def validate_scopes_query(
     """
 
     token_scopes = [Scope.from_str(ts) for ts in token.scopes]
-    scope_intersection = [ts for ts in token_scopes if query_scope.is_superset(ts)]
+    
+    # if query_scope is a point in scope space ("specific"), then we want to
+    # ensure that it sits within the authorized token scopes and return only it;
+    # if it doesn't, then we return an empty list
+    if query_scope.specific():
+        scope_space = ([query_scope] 
+                       if any([Scope.from_str(ts).is_superset(query_scope) for ts in token.scopes]) 
+                       else [])
+
+    # if query_scope is not specific, we want to return all (and only)
+    # authorized token scopes that fall within it
+    else:
+        scopes_space = [ts for ts in token_scopes if query_scope.is_superset(ts)]
+
+
     if as_str:
-        scope_intersection = [str(s) for s in scope_intersection]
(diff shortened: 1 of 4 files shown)


hmacdope:
I should add that another possible way to do this is regex, but I'm not so good at that.

hmacdope:
@dotsdl this is probably ready for a first look.

fah_alchemy/base/api.py
```diff
@@ -79,7 +83,9 @@ def validate_scopes_query(
         ):
 
             # match (query_org == token_org) then (query_campaign == token_campaign) then (query_proj == token_proj)
-            if not (query_field == target_field or query_field is None):
+            if not (
+                query_field == target_field or query_field is None or query_field == "*"
```
dotsdl, line 87: We don't need to test for `'*'` here, as the field will actually be `None` if it's the wildcard value. We use `'*'` in the string representation, but in a `Scope` object this is actually represented as `None`.

This ends up having nice consequences when we use `Scope` objects to query Neo4j, since for a property dict keys with values of `None` end up not being included as query params, requiring us to do little to no processing in the query code.

dotsdl: I've changed `Scope._validate_component` to turn '*' into `None` when given as a component on `init`.

fah_alchemy/tests/unit/test_base_api.py
```diff
@@ -15,6 +26,8 @@ def tokendata():
             "org1-campaignA-projectI",
             "org1-campaignB-projectI",
             "org1-campaignB-projectII",
+            "org2-*-*",
+            "org3-campaignA-*",
```
dotsdl, line 30: I added two additional `Scope`s to this user's authorized list. Can you:
1. fix the tests below that these additions broke, and
2. modify `base.api.validate_scopes` and `base.api.validate_scopes_query` such that they handle the presence of non-specific scopes such as these among a user's authorized set?
    - for `validate_scopes`, though `scope` may be specific, the `Scope`s in `token` won't necessarily be.
    - for `validate_scopes_query`, neither the `query_scope` or the `Scope`s in `token` will necessarily be specific.
    - for both these cases, it may make sense to add a `superset` method or similar to `Scope` that you can use for evaluating `Scope`s against each other. I've added a stub for this.

dotsdl (changes requested):
Great work @hmacdope! I've added comments, along with some direction as to what I think we need yet. 

Now that we have hierarchical handling of components in `Scope`, we want to be able to add `Scope`s with wildcards to a user's authorized list. This means that any `Scope`-based queries or gets a user does with `FahAlchemyClient` must be evaluated against this list to ensure they *can* access the things they are asking for. This is tricky, but I think we're close.

fah_alchemy/base/api.py
hmacdope: Okay makes sense to me to not to have "*" and instead cast to `None`

fah_alchemy/tests/unit/test_base_api.py
hmacdope: @dotsdl I think this has the correct behaviour now. Feel free to take a look. :)

dotsdl (approved):
Great work @hmacdope! Feel free to merge!

PR #70, opened 2023-01-23
[WIP] Scope cli points

### LINKED ISSUE #69: Create CLI points for adding, removing, and listing authorized `Scope`s for a given identity
We can now add, remove, and list identities via `fah-alchemy identity add|remove|list`, but we also need to be able to add, remove, and list `Scope`s that a given identity can interact with.

There are currently stubs for each of these CLI points, but they need implementations.

hmacdope:
Fixes #69 

Currently WIP. 

Adds cli points for adding removing and listing scopes for a user.  

- [ ] Need to fix up scope removal on `statestore`, is failing for wildcards.

### DIFF
--- a/fah_alchemy/storage/statestore.py
+++ b/fah_alchemy/storage/statestore.py
@@ -1238,10 +1238,13 @@ class Neo4jStore(FahAlchemyStateStore):
         with self.transaction() as tx:
             tx.run(q)
 
-    def list_scopes(self, cls: type[CredentialedEntity]) -> List[Scope]:
+    def list_scopes(
+        self, identifier: str, cls: type[CredentialedEntity]
+    ) -> List[Scope]:
         """List all scopes for which the given entity has access."""
         q = f"""
-        MATCH (n:{cls.__name__})-[:HAS_ACCESS]->(s:Scope)
+        MATCH (n:{cls.__name__} {{identifier: '{identifier}'}})
+        MERGE (n)-[r:HAS_SCOPE]->(s:Scope)
         RETURN s
         """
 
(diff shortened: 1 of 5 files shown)


fah_alchemy/storage/statestore.py
```diff
@@ -1237,3 +1237,50 @@ def remove_credentialed_identity(
...
+            scope_rec = record["s"]
+            # use Scope.from_str_tuple to ensure that the scope wildcards are
+            # deserialized properly ie that they are converted from "*" to None
+            scope = Scope.from_str_tuple(
+                (scope_rec["org"], scope_rec["campaign"], scope_rec["project"])
+            )
+            scopes.append(scope)
+        return scopes
+
+    def add_scope(self, identifier: str, cls: type[CredentialedEntity], scope: Scope):
```
dotsdl, line 1265: Instead of Scopes as nodes, let's just add to the `"scopes"` property of the identity node. You'll need to add the given scope to the existing list of properties using e.g.: https://stackoverflow.com/questions/21979782/how-to-push-values-to-property-array-in-cypher#21980719

dotsdl (changes requested):
Let's switch from using Scope nodes to keeping them as property array elements. Looking good so far though!

fah_alchemy/storage/statestore.py
hmacdope: Hopefully addressed this. :)

hmacdope:
@dotsdl I think this is ready for another look. Hopefully I have covered the points we discussed. 😄

dotsdl (approved):
Brilliant @hmacdope! Feel free to merge whenever you're ready!

PR #82, opened 2023-02-03
[DNM] explain how to setup and install the cloudwatch agent + add logging 

mikemhenry:
I am going to also set it up so we can view the container logs on CloudWatch (instead of the current method which is to ssh into the host and attach to the running tmux session)

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -60,6 +60,7 @@ jobs:
         uses: docker/build-push-action@ad44023a93711e3deb337508980b4b5e9bcdc5dc
         with:
           context: .
+          platforms: linux/amd64,linux/arm64
           file: docker/Dockerfile
           push: true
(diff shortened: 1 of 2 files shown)


mikemhenry:
Okay and I am also going to try and build and `arm64v8` image so we can save a ton of money by using a graviton ec2 instance. First I want to see what I can build multi arch wise but CI will probably be really slow, it should be faster when I add a `platform=amd64,arm64v8` arg

dotsdl:
Awesome, thanks for investigating this @mikemhenry!

mikemhenry:
looks like amber tools will be a problem. I am working on getting an arm build of neobolt on conda-forge since we will need that as well (if we want to run on ARM EC2 instances)
https://github.com/conda-forge/neobolt-feedstock/pull/19

mikemhenry:
Will need to get ambertools building on ARM before this will work

PR #90, opened 2023-02-14
Switch ProtocolDAGResultRef to use ok instead of success property

### LINKED ISSUE #89: Switch `ProtocolDAGResultRef` to use `ok` instead of `success` property
Since `ProtocolDAGResultRef.success` carries the same information as its corresponding `ProtocolDAGResult`'s `ok()` method, we might as well give these the same name to avoid confusion.

hmacdope:
Fixes #89 

This was an easy `grep` job so just thought I would bash it out.


dotsdl:
Awesome, thank you @hmacdope!

PR #91, opened 2023-02-16
Add list scopes methods to client and endpoints to API

### LINKED ISSUE #83: Add `list_scopes` method to `AlchemiscaleClient`
Users should be able to interrogate what scopes they have access to via the `AlchemiscaleClient. We would like a `list_scopes` method on the client that returns the contents of their user identity's `scopes` list.

hmacdope:
Fixes #83 

- Currently returns the scopes as strings rather than `Scope` models, I thought this was more user-friendly than the other way round.


hmacdope:
@dotsdl this should be ready for review now :)

alchemiscale/compute/api.py
```diff
@@ -76,6 +80,17 @@ async def check(
     _check_store_connectivity(n4js, s3os)
 
 
+@router.get("/identities/{identity_identifier}/scopes")
```
dotsdl, line 83: The typical pattern for RESTful APIs is to go `/<entity_plural>/<entity_identifier>/`, with anything following being used to grab something specific for that entity, such as `scopes` like we do here. We use this pattern on other endpoints for this API as well, so I've made this change here to reflect that.

alchemiscale/compute/client.py
```diff
@@ -31,6 +31,13 @@ class AlchemiscaleComputeClient(AlchemiscaleBaseClient):
 
     _exception = AlchemiscaleComputeClientError
 
+    def list_scopes(self) -> List[Scope]:
+        scopes = self._get_resource(
+            f"/identities/{self.identifier}/scopes",
+            return_gufe=False,
+        )
+        return [Scope.from_str(s) for s in scopes]
```
dotsdl, line 39: I think it's more consistent for us to return `Scope` objects, since these can be directly used where a `Scope` is needed elsewhere, and it doesn't impact readability of printed results in my view.

dotsdl (approved):
Great work @hmacdope! I've made some alterations, with comments. Ready to merge when you are satisfied.

PR #93, opened 2023-02-20
Added cloudwatch logging

mikemhenry:


mikemhenry:
Few different ways to organize this, I made one log group and set up a stream for each container, with the service name and the container ID, so that should help us track down the right log we want to look at. Also set things up for a 30 day retention. 
image

mikemhenry:
When you download logs, the file that gets created is called `log-events-viewer-result.csv` so no worries with the slashes

mikemhenry:
Updated screenshot:

image


This is what the logs look like after an update, so you have see how service names are duplicated, but container IDs are unique. I also added more columns so you can see start times as well.

dotsdl:
Reviewing!

docker/docker-compose.yml
```diff
@@ -44,6 +44,12 @@ services:
     user: ${USER_ID:?err}:${GROUP_ID:?err}
     labels:
       - "traefik.enable=false"
+    logging:
+       driver: "awslogs"
+       options:
+          awslogs-region: "us-east-1"
```
dotsdl, line 50:
```suggestion
          awslogs-region: "${AWS_DEFAULT_REGION}"
```

docker/docker-compose.yml
```diff
@@ -82,6 +88,12 @@ services:
       timeout: 10s
       retries: 3
       start_period: 20s
+    logging:
+       driver: "awslogs"
+       options:
+          awslogs-region: "us-east-1"
```
dotsdl, line 94:
```suggestion
          awslogs-region: "us-east-1"
```

docker/docker-compose.yml
```diff
@@ -122,6 +134,12 @@ services:
...
```
dotsdl, line 140:
```suggestion
          awslogs-region: "${AWS_DEFAULT_REGION}"
```

docker/docker-compose.yml
```diff
@@ -144,6 +162,12 @@ services:
       neo4j:
         condition: service_healthy
     command: "database init"
+    logging:
+       driver: "awslogs"
+       options:
+          awslogs-region: "us-east-1"
```
dotsdl, line 168:
```suggestion
          awslogs-region: "${AWS_DEFAULT_REGION}"
```

docker/docker-compose.yml
```diff
@@ -169,3 +193,9 @@ services:
       # So that Traefik can listen to the Docker events
       - /var/run/docker.sock:/var/run/docker.sock
       - ./letsencrypt:/letsencrypt
+    logging:
+       driver: "awslogs"
+       options:
+          awslogs-region: "us-east-1"
```
dotsdl, line 199:
```suggestion
          awslogs-region: "${AWS_DEFAULT_REGION}"
```

dotsdl (approved):
Looks great @mikemhenry! Just one suggestion if you think it will work.

docker/docker-compose.yml
dotsdl: Sorry, this was meant to be:

```suggestion
          awslogs-region: "${AWS_DEFAULT_REGION}"
```

mikemhenry:
I do want to set it to error if unset, but great catch!

mikemhenry:
@dotsdl forgot about making the change! Will do asap

PR #95, opened 2023-02-27
Task lifecycle implementation

### LINKED ISSUE #85: Design Task status lifecycle; implement
@hmacdope and I worked out a proposed status lifecycle for tasks, with the following statuses:
- `waiting` : default status on creation; eligible for claiming if actioned on a `TaskHub`
- `running` : status once claimed; indicates task is being executed by a compute service; only `waiting` tasks can become `running` tasks; `running` tasks can be returned to `waiting` by the compute service that claimed them (such as on shutdown) or by the `ComputeAPIService` via heartbeat expiration
- `complete` : once executed and with `ProtocolDAGResult.ok() == True`, indicates the task is complete; cannot be returned to `waiting`
- `error` : once executed and with `ProtocolDAGResult.ok() == False`, indicates task has errored; can be returned to `waiting` to be re-run
- `invalid` : used to mark a task as being flawed in some way without "deletion"; marked by user, can be reached from any other status; an `invalid` task cannot be set to any other status thereafter
- `deleted` : used to "delete" a task; marked by user, can be reached from any other status; a `deleted` task cannot be set to any other status thereafter

Additional notes:
1. If a compute service finishes a Task, but its status has been set to anything other than 'running', then the service will not set the status at all.

2. Only `running` Tasks should have CLAIMS relationship (see (3) below).

3. Consider a CLAIMS relationship to a ComputeServiceIdentity node from a Task that is claimed instead of `claim` property

4. Currently, `invalid` and `deleted` statuses cannot be undone; or rather, the model still allows it, but we don't expose a way to do this.

5. Task "ACTIONS" relationships are independent of status; however, only `waiting` tasks that do not "EXTENDS" a non `complete` Task can be claimed.

6. We will consider removing all "ACTIONS" relationships from a Task that is marked `complete`, `invalid`, or `deleted`; the latter two are terminal statuses, while `complete` can only become one of `invalid` or `deleted`; none of these can return to `waiting`. This will declutter TaskHub relationships; there is little reason to keep them once these statuses are reached.

7. It should not be possible to create a Task that `EXTENDS` an `invalid` or `deleted` Task; there is little harm in allowing it, but little reason to either

8. If a Task is marked either `invalid` or `deleted`, then any Task **that EXTENDS it** is recursively marked with that status as well.

See attached diagram for these and other details.

alchemiscale - task lifecycle drawio(1)

hmacdope:
Fixes #85 

Implementation of task lifecycle. 

TODO:

- [x] Check design with @dotsdl 
- [x] Check `set_status` vs `set_complete`. 
~~- [ ] Add `ComputeServiceIdentity` `CLAIMS` relationships~~ This is to be handled in another PR, possibly #98.

### DIFF
--- a/alchemiscale/compute/api.py
+++ b/alchemiscale/compute/api.py
@@ -211,7 +211,7 @@ async def set_task_status(
     token: TokenData = Depends(get_token_data_depends),
 ):
     """
-    NOTE: The strict_complete=False kwarg (the default case regardless) is to
+    NOTE: The raise_error=False kwarg (the default case regardless) is to
     remind us that we are choosing the no-op behavior, rather than raising an
     error when using `set_task_complete` to attempt to finish a task that has
     been removed from running. This has no effect on the behavior of other
@@ -220,7 +220,7 @@ async def set_task_status(
     task_sk = ScopedKey.from_str(task_scoped_key)
     validate_scopes(task_sk.scope, token)
     status = TaskStatusEnum(status)
-    n4js.set_task_status([task_sk], status, strict_complete=False)
+    n4js.set_task_status([task_sk], status, raise_error=False)
 
 
 @router.get("/tasks/{task_scoped_key}/status")
--- a/alchemiscale/compute/api.py
+++ b/alchemiscale/compute/api.py
@@ -219,6 +219,7 @@ async def set_task_status(
     """
     task_sk = ScopedKey.from_str(task_scoped_key)
     validate_scopes(task_sk.scope, token)
+
     status = TaskStatusEnum(status)
     n4js.set_task_status([task_sk], status, raise_error=False)
 
@@ -232,6 +233,7 @@ async def get_task_status(
 ):
     task_sk = ScopedKey.from_str(task_scoped_key)
     validate_scopes(task_sk.scope, token)
+
     status = n4js.get_task_status([task_sk])
     # cast the Python Enum to a string, as the JSON encoder doesn't know how
     # to handle it
--- a/alchemiscale/compute/api.py
+++ b/alchemiscale/compute/api.py
@@ -202,6 +202,19 @@ def set_task_result(
     return result_sk
 
 
+@router.post("/tasks/{task_scoped_key}/status")
+def set_task_status(
+    task_scoped_key,
+    status: str = Body(),
+    n4js: Neo4jStore = Depends(get_n4js_depends),
+    token: TokenData = Depends(get_token_data_depends),
+):
+    task_sk = ScopedKey.from_str(task_scoped_key)
+    validate_scopes(task_sk.scope, token)
+    status = TaskStatusEnum(status)
+    n4js.set_task_status(task_sk, status)
+
+
 @router.get("/chemicalsystems")
 async def chemicalsystems():
     return {"message": "nothing yet"}
--- a/alchemiscale/compute/api.py
+++ b/alchemiscale/compute/api.py
@@ -214,7 +214,8 @@ async def set_task_status(
     validate_scopes(task_sk.scope, token)
 
     status = TaskStatusEnum(status)
-    n4js.set_task_status([task_sk], status)
+    tasks_statused = n4js.set_task_status([task_sk], status)
+    return [str(t) if t is not None else None for t in tasks_statused][0]
 
 
 @router.get("/tasks/{task_scoped_key}/status")
--- a/alchemiscale/compute/api.py
+++ b/alchemiscale/compute/api.py
@@ -210,18 +210,11 @@ async def set_task_status(
     n4js: Neo4jStore = Depends(get_n4js_depends),
     token: TokenData = Depends(get_token_data_depends),
 ):
-    """
-    NOTE: The raise_error=False kwarg (the default case regardless) is to
-    remind us that we are choosing the no-op behavior, rather than raising an
-    error when using `set_task_complete` to attempt to finish a task that has
-    been removed from running. This has no effect on the behavior of other
-    `set_task_status` methods.
-    """
     task_sk = ScopedKey.from_str(task_scoped_key)
     validate_scopes(task_sk.scope, token)
 
     status = TaskStatusEnum(status)
-    n4js.set_task_status([task_sk], status, raise_error=False)
+    n4js.set_task_status([task_sk], status)
 
 
 @router.get("/tasks/{task_scoped_key}/status")
@@ -235,9 +228,8 @@ async def get_task_status(
     validate_scopes(task_sk.scope, token)
 
     status = n4js.get_task_status([task_sk])
-    # cast the Python Enum to a string, as the JSON encoder doesn't know how
-    # to handle it
-    return status[task_sk].value
+
+    return status[0].value
 
 
 @router.get("/chemicalsystems")
--- a/alchemiscale/compute/api.py
+++ b/alchemiscale/compute/api.py
@@ -209,10 +209,17 @@ async def set_task_status(
     n4js: Neo4jStore = Depends(get_n4js_depends),
     token: TokenData = Depends(get_token_data_depends),
 ):
+    """
+    NOTE: The strict_complete=False kwarg (the default case regardless) is to
+    remind us that we are choosing the no-op behavior, rather than raising an
+    error when using `set_task_complete` to attempt to finish a task that has
+    been removed from running. This has no effect on the behavior of other
+    `set_task_status` methods.
+    """
     task_sk = ScopedKey.from_str(task_scoped_key)
     validate_scopes(task_sk.scope, token)
     status = TaskStatusEnum(status)
-    n4js.set_task_status([task_sk], status)
+    n4js.set_task_status([task_sk], status, strict_complete=False)
 
 
 @router.get("/tasks/{task_scoped_key}/status")
--- a/alchemiscale/compute/api.py
+++ b/alchemiscale/compute/api.py
@@ -203,7 +203,7 @@ def set_task_result(
 
 
 @router.post("/tasks/{task_scoped_key}/status")
-def set_task_status(
+async def set_task_status(
     task_scoped_key,
     status: str = Body(),
     n4js: Neo4jStore = Depends(get_n4js_depends),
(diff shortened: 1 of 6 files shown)


alchemiscale/compute/api.py
```diff
@@ -202,6 +202,31 @@ def set_task_result(
     return result_sk
 
 
+@router.post("/tasks/{task_scoped_key}/status")
```
hmacdope, line 205: Reminder to have API to enforce checks for state transitions.

alchemiscale/compute/client.py
```diff
@@ -95,3 +95,31 @@ def set_task_result(
...
+
+        self._post_resource(f"tasks/{task}/status", status.value)
+
+    def get_task_status(self, task: ScopedKey) -> TaskStatusEnum:
+        """Get the status of a `Task`."""
+        return TaskStatusEnum(
+            self._get_resource(f"tasks/{task}/status", return_gufe=False)
+        )
+
+    def set_task_complete(self, task: ScopedKey) -> None:
```
hmacdope, line 114: Race condition issue, this should be atomic. Make this a single transaction cypher query on the compute API side

alchemiscale/storage/statestore.py
```diff
@@ -1452,56 +1458,252 @@ def get_task_failures(self, task: ScopedKey) -> List[ProtocolDAGResultRef]:
...
-        MATCH (t:Task {{_scoped_key: "{task}"}})
-        SET t.status = 'waiting'
+    def set_task_status(self, task: ScopedKey, status: TaskStatusEnum) -> ScopedKey:
+        """
+        Set the status of a task, this is a master method that calls the
+        appropriate method for the status.
+
+        Parameters
+        ----------
+        task : Union[ScopedKey,List[ScopedKey]]
```
hmacdope, line 1468: Fix type hint

alchemiscale/storage/statestore.py
```diff
@@ -1452,56 +1458,252 @@ def get_task_failures(self, task: ScopedKey) -> List[ProtocolDAGResultRef]:
...
+                    raise ValueError(
+                        f"Cannot set task {t} with current status: {status} to `running` as it is not currently `waiting`."
+                    )
+                q2 = f"""
+                MATCH (t:Task {{_scoped_key: '{t}'}})
+                SET t.status = '{TaskStatusEnum.running.value}'
+                """
+                tx.run(q2)
+
+    def set_task_complete(self, task: Union[ScopedKey, List[ScopedKey]]) -> None:
```
hmacdope, line 1549: Make all of these a list only rather than `Union[ScopedKey, List[ScopedKey]]`

alchemiscale/storage/statestore.py
```diff
@@ -1452,56 +1458,252 @@ def get_task_failures(self, task: ScopedKey) -> List[ProtocolDAGResultRef]:
...
+        """
+        if isinstance(task, ScopedKey):
+            task = [task]
+
+        with self.transaction() as tx:
+            for t in task:
+                # set the status and delete any ACTIONS relationship
+                q = f"""
+                MATCH (t:Task {{_scoped_key: '{t}'}})
+                SET t.status = '{TaskStatusEnum.deleted.value}'
```
hmacdope, line 1701: Can simplify due to string `__repr__` casting

alchemiscale/storage/statestore.py
```diff
@@ -1452,56 +1458,252 @@ def get_task_failures(self, task: ScopedKey) -> List[ProtocolDAGResultRef]:
...
+        Set the status of a task or list of tasks to `invalid`.
+
+        As per the design of the `Task` data lifecycle any task can be set to `invalid`.
+        Once `invalid` is set the task cannot change status
+        """
+        if isinstance(task, ScopedKey):
+            task = [task]
+
+        with self.transaction() as tx:
+            for t in task:
```
hmacdope, line 1672: Make sure EXTENDS chain is also marked invalid.

alchemiscale/storage/statestore.py
```diff
@@ -1452,56 +1458,252 @@ def get_task_failures(self, task: ScopedKey) -> List[ProtocolDAGResultRef]:
...
+        Once `deleted` is set the task cannot change status
+        """
+        if isinstance(task, ScopedKey):
+            task = [task]
+
+        with self.transaction() as tx:
+            for t in task:
+                # set the status and delete any ACTIONS relationship
+                q = f"""
+                MATCH (t:Task {{_scoped_key: '{t}'}})
```
hmacdope, line 1700: Make sure EXTENDS chain is also marked invalid and EXTENDS tasks also have their ACTIONS relationship removed.

alchemiscale/tests/integration/storage/test_statestore.py
```diff
@@ -1245,3 +1272,469 @@ def test_remove_scope_duplicate(
...
+    def f_set_task_deleted(self, n4js):
+        return n4js.set_task_deleted
+
+    # allowed = [running, error, invalid, deleted]
+    # no-op = [waiting]
+    # not allowed = [complete]
+    @pytest.mark.parametrize(
+        "status_func, status",
+        [
+            ("f_set_task_running", TaskStatusEnum.running),
```
hmacdope, line 1310: Add in `failure_flag` to avoid specific pytest raises cases and  repeat testing of failures,

dotsdl (changes requested):
Looking great @hmacdope! See comments from our discussion today; just a few changes needed but we're 90% there.

alchemiscale/storage/statestore.py
hmacdope: This doesn't work as the __repr__ of `TaskStatusEnum.deleted` is `TaskStatusEnum.deleted` rather than `deleted`

hmacdope:
@dotsdl this is probably ready for another reveiw! 

The only main design decision I have introduced since we last spoke is that for `set_task_complete` one can either choose :

-  `strict_complete=True`,  task not running = `Exception` 
-  `strict_complete=False` task not running = `no-op` 

behaviour. This is to allow both atomic-like (demand that you are the person to set the running to complete transition) and non-atomic (if someone else set the status before us that is ok, we do nothing) behaviour when completing a task. I thought this might be useful down the road, but am happy to remove if you don't think its necessary, or unnecessarily complex. The client side calls use the `strict-complete=False` behaviour as someone else can mark the task before it is complete. 

Let me know either way :)

alchemiscale/storage/statestore.py
```diff
@@ -1463,56 +1469,257 @@ def get_task_failures(self, task: ScopedKey) -> List[ProtocolDAGResultRef]:
...
+                tx.run(q2)
 
     def set_task_complete(
-        self,
-        task: ScopedKey,
-    ):
-        q = f"""
-        MATCH (t:Task {{_scoped_key: "{task}"}})
-        SET t.status = 'complete'
+        self, tasks: List[ScopedKey], strict_complete: Optional[bool] = False, **kwargs
```
hmacdope, line 1557: change to raise_error

hmacdope: done

hmacdope:
@dotsdl this should be ready for a final review. :)

dotsdl (approved):
@hmacdope I made some major changes to the setters, motivated in part to avoid potential race conditions by performing status checking in separate queries from setting. I then made the behavior of each setter follow the same pattern, with `raise_error=False` as a kwarg. 

The default behavior for each of these is that they will return the `ScopedKey` of each `Task` that is set to the desired status, `None` for those cases where this couldn't be done. This is consistent with how we've got similar methods working, such as `action_tasks`.

dotsdl:
@hmacdope since my changes are pretty substantial, please let me know if you have any reservations about me merging. I'll wait for word from you before proceeding.

hmacdope (commented):
@dotsdl few quick comments.

alchemiscale/storage/statestore.py
```diff
@@ -1463,56 +1469,260 @@ def get_task_failures(self, task: ScopedKey) -> List[ProtocolDAGResultRef]:
...
-    def set_task_running(self, task: ScopedKey, computekey: ComputeKey):
-        ...
+        return statuses
+
+    def _set_task_status(self, tasks, q_func, err_msg_func, raise_error):
+        tasks_set = []
+        with self.transaction() as tx:
+            for t in tasks:
+                res = tx.run(q_func(t))
+                for record in res:
```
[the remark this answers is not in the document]
hmacdope: If its not possible to have more than one record by design, perhaps leave a comment or use some other syntax  other than a `for` loop that makes it clearer?

alchemiscale/compute/api.py
```diff
@@ -203,6 +203,35 @@ def set_task_result(
...
+    task_scoped_key,
+    status: str = Body(),
+    n4js: Neo4jStore = Depends(get_n4js_depends),
+    token: TokenData = Depends(get_token_data_depends),
+):
+    task_sk = ScopedKey.from_str(task_scoped_key)
+    validate_scopes(task_sk.scope, token)
+
+    status = TaskStatusEnum(status)
+    n4js.set_task_status([task_sk], status)
```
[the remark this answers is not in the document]
dotsdl: Done!

alchemiscale/compute/client.py
```diff
@@ -94,3 +94,12 @@ def set_task_result(
         pdr_sk = self._post_resource(f"tasks/{task}/results", data)
 
         return ScopedKey.from_dict(pdr_sk)
+
+    def set_task_status(self, task: ScopedKey, status: TaskStatusEnum) -> None:
+        """Set the status of a `Task`."""
+        self._post_resource(f"tasks/{task}/status", status.value)
```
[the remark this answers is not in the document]
dotsdl: Done!

alchemiscale/storage/statestore.py
```diff
@@ -1463,56 +1469,260 @@ def get_task_failures(self, task: ScopedKey) -> List[ProtocolDAGResultRef]:
...
+                else:
+                    status = task.get("status")
+                    statuses.append(TaskStatusEnum(status))
 
-    def set_task_running(self, task: ScopedKey, computekey: ComputeKey):
-        ...
+        return statuses
+
+    def _set_task_status(self, tasks, q_func, err_msg_func, raise_error):
+        tasks_set = []
```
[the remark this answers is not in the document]
dotsdl: Agreed it's ambiguous; using `set` as a verb here. Changed it.

alchemiscale/storage/statestore.py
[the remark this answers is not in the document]
dotsdl: Ah, good point; I added an exception that will get raised if more than one record shows up. It should not be possible.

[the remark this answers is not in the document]
dotsdl: Went for a `break` and a comment instead. Under certain conditions we will get more than one record back; I believe it has to do with whether any additional `OPTIONAL MATCH` or `MATCH` statements end up triggering. Because of the constraint on these objects, however, we will never have actual duplicate nodes in the DB.

PR #97, opened 2023-03-04
Add support for `Settings` in `Neo4jStore` (fixes for latest `gufe`)

dotsdl:
- Added `Settings` support to state store
- Fixes to allow for presence of Settings in serialized/deserialized bits

### DIFF
--- a/alchemiscale/base/client.py
+++ b/alchemiscale/base/client.py
@@ -17,6 +17,10 @@ from ..models import Scope, ScopedKey
 from ..storage.models import TaskHub, Task
 
 
+def json_to_gufe(jsondata):
+    return GufeTokenizable.from_dict(json.loads(jsondata, cls=JSON_HANDLER.decoder))
+
+
 class AlchemiscaleBaseClientError(Exception):
     ...
 
@@ -81,14 +85,14 @@ class AlchemiscaleBaseClient:
 
         if params.get("return_gufe"):
             return {
-                ScopedKey.from_str(k): GufeTokenizable.from_dict(v)
+                ScopedKey.from_str(k): json_to_gufe(v)
                 for k, v in resp.json().items()
             }
         else:
             return [ScopedKey.from_str(i) for i in resp.json()]
 
     @_use_token
-    def _get_resource(self, resource, params=None, return_gufe=True):
+    def _get_resource(self, resource, params=None):
         if params is None:
             params = {}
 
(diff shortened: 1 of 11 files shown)


dotsdl:
Ready for review! @mikemhenry and @hmacdope, whichever one of you wants to review, I'd be most grateful!

alchemiscale/base/client.py
```diff
@@ -119,9 +118,9 @@ def _post_resource(self, resource, data):
         return resp.json()
 
     def get_info(self):
-        return self._get_resource("/info", params={}, return_gufe=False)
+        return self._get_resource("/info")
 
     @_use_token
     def _api_check(self):
         # Check if the API is up and running and can reach services
-        self._get_resource("/check", params={}, return_gufe=False)
```
hmacdope, line 127: I think this is a good idea moving the gufe logic outside this method, I kept getting caught out by it.

alchemiscale/storage/statestore.py
```diff
@@ -319,6 +320,11 @@ def _gufe_to_subgraph(
                 ):
                     node[key] = json.dumps(value, cls=JSON_HANDLER.encoder)
                     node["_json_props"].append(key)
+            elif isinstance(value, Settings):
+                # TODO: finish up approach here for serializing settings
```
hmacdope, line 324: Is this still TODO?

alchemiscale/storage/statestore.py
```diff
@@ -1358,7 +1364,10 @@ def get_task_transformation(
         self,
         task: ScopedKey,
         return_gufe=True,
-    ) -> Tuple[Transformation, Optional[ProtocolDAGResultRef]]:
+    ) -> Union[
+        Tuple[Transformation, Optional[ProtocolDAGResultRef]],
```
hmacdope, line 1368: Given the two possible return types might be good to have a quick note in docstring as to under which circumstances each is returned

hmacdope (changes requested):
Just a few nitpicks, but looking great! Thanks for fixing up.

alchemiscale/base/client.py
dotsdl: Yeah, we ended up disabling the behavior 90% of the time, so seemed right to just drop it.

alchemiscale/storage/statestore.py
dotsdl: Nope! All accounted for. Thanks for catching this!

alchemiscale/storage/statestore.py
dotsdl: Done!

dotsdl:
Comments addressed @hmacdope! Thanks for the review!

hmacdope (approved):
Thanks @dotsdl

PR #98, opened 2023-03-04
Add missing features, polish to `SynchronousComputeService`

### LINKED ISSUE #34: Compute : `SynchronousComputeService` implementation with core features
We require a compute service that serves as the reference implementation for any compute service that interacts with the `FahAlchemyComputeAPI` to pull tasks, execute `ProtocolDAG`s, and push results back. This reference implementation should perform all activity in-process, in-serial, for ease of testing and debugging.

More performance-focused compute services will build on the components implemented in this reference implementation.

dotsdl:
Closes #34.


alchemiscale/compute/service.py
```diff
@@ -178,62 +234,94 @@ def execute(self, task: ScopedKey) -> ScopedKey:
...
+    def cycle(self, task_limit):
+        if task_limit is not None:
+            if self.counter >= task_limit:
+                self.logger.info("Performed %s tasks; beyond task limit %s", self.counter, task_limit)
+                return
+
+        # claim tasks from the compute API
+        self.logger.info("Claiming tasks")
+        tasks: List[ScopedKey] = self.claim_tasks(self.limit)
+        self.logger.info("Claimed %d tasks", len([t for t in tasks if t is not None]))
```
hmacdope, line 262: I would also dump the scoped keys to the log?

hmacdope (changes requested):
Few little queries but overall looking good.

dotsdl:
@mikemhenry do you have a favorite logging formatter you'd like us to use for compute service logs?

alchemiscale/compute/service.py
dotsdl: We already indicate which task we execute and finish further below. I think the log may be too noisy doing it here as well.

alchemiscale/compute/api.py
```diff
@@ -92,6 +98,42 @@ async def list_scopes(
...
+async def register_computeservice(
+    compute_service_id,
+    n4js: Neo4jStore = Depends(get_n4js_depends),
+):
+    now = datetime.utcnow()
+    csreg = ComputeServiceRegistration(
+        identifier=compute_service_id, registered=now, heartbeat=now
+    )
+
+    n4js.register_computeservice(csreg)
```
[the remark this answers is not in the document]
dotsdl: Done!

dotsdl:
Ready for final review @hmacdope!

alchemiscale/compute/api.py
```diff
@@ -92,6 +97,53 @@ async def list_scopes(
...
+    settings: ComputeAPISettings = Depends(get_base_api_settings),
+):
+    now = datetime.utcnow()
+
+    # expire any stale registrations, along with their claims
+    expire_delta = timedelta(
+        seconds=settings.ALCHEMISCALE_COMPUTE_API_REGISTRATION_EXPIRE_SECONDS
+    )
+    expire_time = now - expire_delta
+    n4js.expire_registrations(expire_time)
```
hmacdope, line 140: Query: is there any need for this to return or expose the expired `compute_service_id`s? I'll leave it up to you, im sure this is fine for now.

alchemiscale/compute/service.py
```diff
@@ -74,59 +78,143 @@ def __init__(
...
-        # claim tasks from the taskhub
-        tasks = self.client.claim_taskhub_tasks(
-            taskhub, claimant=self.name, count=count
-        )
+        # claim tasks from taskhubs based on weight; keep going till we hit our
+        # total desired task count, or we run out of taskhubs to draw from
+        while len(tasks) < count and len(taskhubs) > 0:
+            # based on weights, choose taskhub to draw from
+            taskhub: List[ScopedKey] = random.choices(
+                list(taskhubs.keys()), weights=[tq.weight for tq in taskhubs.values()]
```
hmacdope, line 201: nitpick: th clearer than tq
```suggestion
                list(taskhubs.keys()), weights=[th.weight for th in taskhubs.values()]
```

alchemiscale/compute/service.py
```diff
@@ -74,59 +78,143 @@ def __init__(
...
+                count=(count - len(tasks)),
+            )
+
+            # gather up claimed tasks, if present
+            for t in claimed_tasks:
+                if t is not None:
+                    tasks.append(t)
+
+            # remove this taskhub from the options available; repeat
+            taskhubs.pop(taskhub)
```
hmacdope, line 217: Im not quite sure I understand this pop, what if we didn't take all the tasks from the hub? I'm probably missing something.

alchemiscale/storage/statestore.py
```diff
@@ -91,24 +92,35 @@ def _select_task_from_taskpool(taskpool: Subgraph) -> Union[ScopedKey, None]:
...
-    SET t.status = 'running', t.claim = '{claimant}'
+    WHERE NOT (t)<-[:CLAIMS]-(:ComputeServiceRegistration)
+    SET t.status = 'running'
+
+    WITH t
+
+    // create CLAIMS relationship with given compute service
+    MATCH (csreg:ComputeServiceRegistration {{identifier: '{compute_service_id}'}})
+    CREATE (t)<-[cl:CLAIMS {{claimed: localdatetime('{datetime.utcnow().isoformat()}')}}]-(csreg)
+
```
hmacdope, line 123: This looks good!

alchemiscale/storage/statestore.py
```diff
@@ -730,6 +745,96 @@ def set_strategy(
...
+        Tasks to avoid collisions in Task execution.
+
+        """
+
+        node = Node(
+            "ComputeServiceRegistration", **compute_service_registration.to_dict()
+        )
+
+        with self.transaction() as tx:
+            tx.merge(
```
hmacdope, line 764: Does this merge allow you to attempt to register the same `compute_service` more than once? IMO should be an error if you try and overwrite and existing service or at least give some feedback if the query did nothing ie found an already existing service. I could be missing something that prevents this though...

alchemiscale/storage/statestore.py
```diff
@@ -730,6 +745,96 @@ def set_strategy(
...
+        and with status `running` will have their status set to `waiting`.
+
+        """
+        q = f"""
+        MATCH (n:ComputeServiceRegistration {{identifier: '{compute_service_id}'}})
+
+        OPTIONAL MATCH (n)-[cl:CLAIMS]->(t:Task {{status: 'running'}})
+        SET t.status = 'waiting'
+
+        DETACH DELETE n
```
hmacdope, line 789: nitpick but should return `n` from the cypher query and propagate that value if possible.

alchemiscale/storage/statestore.py
```diff
@@ -730,6 +745,96 @@ def set_strategy(
...
+        with self.transaction() as tx:
+            tx.run(q)
+
+        return compute_service_id
+
+    def expire_registrations(self, expire_time: datetime):
+        """Remove all registrations with last heartbeat prior to the given `expire_time`."""
+        q = f"""
+        MATCH (n:ComputeServiceRegistration)
+        WHERE n.heartbeat < localdatetime('{expire_time.isoformat()}')
```
hmacdope, line 816: Cypher does datetime comparison? Cool!

alchemiscale/storage/statestore.py
```diff
@@ -1199,6 +1304,14 @@ def create_task(
 
         """
         scope = transformation.scope
+        if "Transformation" not in transformation.qualname:
+            raise ValueError(
+                "`transformation` ScopedKey does not correspond to a `Transformation`"
+            )
+
+        if extends is not None and extends.qualname != "Task":
```
hmacdope, line 1312: useful check, I'm sure there are other sharp edges of this nature floating around.

alchemiscale/tests/integration/storage/test_statestore.py
```diff
@@ -270,6 +273,92 @@ def test_get_transformation_failures(
...
+        assert csreg["registered"] == now
+        assert csreg["heartbeat"] == now
+
+    def test_deregister_computeservice(self, n4js, compute_service_id):
+        now = datetime.utcnow()
+        registration = ComputeServiceRegistration(
+            identifier=compute_service_id, registered=now, heartbeat=now
+        )
+
+        n4js.register_computeservice(registration)
```
hmacdope, line 301: test return values from database queries here also ?

alchemiscale/tests/integration/storage/test_statestore.py
```diff
@@ -270,6 +273,92 @@ def test_get_transformation_failures(
...
+
+        assert csreg is None
+
+    def test_heartbeat_computeservice(self, n4js, compute_service_id):
+        now = datetime.utcnow()
+        registration = ComputeServiceRegistration(
+            identifier=compute_service_id, registered=now, heartbeat=now
+        )
+
+        n4js.register_computeservice(registration)
```
hmacdope, line 321: Also check return values here?

alchemiscale/tests/integration/test_cli.py
```diff
@@ -136,18 +142,92 @@ def test_compute_api(n4js, s3os):
...
     jwt_opts = []  # leaving empty, we have default behavior for these
 
+    return host, port, (command + api_opts + db_opts + s3_opts + jwt_opts)
+
+
+@pytest.fixture
+def compute_service_config(compute_api_args):
+    host, port, _ = compute_api_args
+
+    config = {
```
hmacdope, line 152: If this YAML is required perhaps in example sometwhere in `devtools` is a good idea?

hmacdope (changes requested):
Looking great @dotsdl, just a few minor things. Ping me again when you want another review, its super close.

alchemiscale/storage/statestore.py
dotsdl: Thank you!

alchemiscale/storage/statestore.py
dotsdl: Good question!

We're using `py2neo`'s `Transaction.merge` here: https://py2neo.org/2021.1/workflow.html?highlight=merge#py2neo.Transaction.merge

It won't create a `ComputeServiceRegistration` with the same `identifier`, and for additional safety on this we also set a uniqueness constraint for `ComputeServiceRegistration`s on their `identifier` (see above). That being said, no feedback will be given if the same `ComputeServiceRegistration` is trying to be used twice; I don't think this will be a problem in practice, but you raise a good point that we also don't want this happening.

I can change this to use `tx.create` instead, which will raise an exception if the object exists already due to the constraint we have set above.

alchemiscale/storage/statestore.py
dotsdl: I don't think we can return `n` itself, since we are deleting it. However, I can use the trick we're using below to return the identifier after extracting it with `WITH` prior to the delete.

alchemiscale/storage/statestore.py
dotsdl: It does! :grin:

alchemiscale/storage/statestore.py
dotsdl: There most certainly are; I tripped on this one myself while doing live testing. I'll sweep through the other methods that are susceptible to this problem and add appropriate checks.

alchemiscale/tests/integration/storage/test_statestore.py
dotsdl: Ah, good catch; will add this.

alchemiscale/tests/integration/storage/test_statestore.py
dotsdl: Will do!

alchemiscale/tests/integration/test_cli.py
dotsdl: Good call, will add an example file. I think this may get better with #107 as well.

alchemiscale/compute/api.py
dotsdl: Not at the moment no. The approach we take here (stacking in expiration with the heartbeat API call) is actually a bit of an inefficient hack; the better approach to this will be to have a separate `PeriodicsService` that performs operations like this on some schedule, since they only have to happen once in a while. This will avoid thrashing the database.

I'll create an issue articulating the vision for this service so we can implement for a future release.

alchemiscale/compute/service.py
dotsdl: Ah, thank you! Yes this is a leftover from the old `TaskQueue` implementation.

alchemiscale/compute/service.py
dotsdl: The idea I've tried to implement here is explained in the comments. I'll try to express it in a different way:
1. When we enter this method, we want to claim `count` `Task`s. 
2. We start by randomly selecting a `TaskHub` to draw from based on their relative weights.
3. We try to draw `count` `Task`s from that `TaskHub`. However, it's possible that this `TaskHub` doesn't actually have enough `Task`s to meet our desired `count`. So, we drop it from the set we select from and roll the dice again, repeating from (2).
4. We keep going until we have either collected `count` `Task`s or run out of `TaskHub`s to draw `Task`s from.

Does the implementation above accomplish this, in your view?

alchemiscale/compute/api.py
dotsdl: Issue here: https://github.com/openforcefield/alchemiscale/issues/109

alchemiscale/storage/statestore.py
dotsdl: I think I've filled many of the similar potholes in.

dotsdl:
Changes complete @hmacdope! Ready for another review!

hmacdope (approved):
You satisfied all my queries @dotsdl! Feel free to merge when ready.

dotsdl:
Thanks @hmacdope! This PR is much better following your reviews. Merging!

PR #99, opened 2023-03-06
Enforce alphanumeric, underscore, or only asterisk for Scope components 

### LINKED ISSUE #88: Enforce alphanumeric, underscore, or only asterisk for `Scope` components
In our `Scope` model, we currently validate that the components `org`, `campaign`, and `project` don't contain dashes `-` or `*`, though `*` alone is valid. However, we really only want each of these components to:
- contain alphanumerics
- contain possibly underscores
- be a single asterisk `*`

We should enforce these instead, perhaps via regex.

hmacdope:
Fixes #88 

Uses a regex to check that scope components are what we expect them to be.

### DIFF
--- a/alchemiscale/models.py
+++ b/alchemiscale/models.py
@@ -6,6 +6,7 @@ Data models --- :mod:`alchemiscale.models`
 from typing import Optional, Union
 from pydantic import BaseModel, Field, validator, root_validator
 from gufe.tokenization import GufeKey
+from re import fullmatch
 
 
 class Scope(BaseModel):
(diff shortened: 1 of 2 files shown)


dotsdl (approved):
Looks great @hmacdope! I made some adjustments, in particular requiring that `Scope` components start with a letter (cannot start with a `_` or number). This is to avoid users creating scopes like `99-22-33` or `_-_-_`, which can get a bit unreadable and a bit odd.

Do you agree?

hmacdope:
@dotsdl sounds good to me! Feel free to merge.

PR #100, opened 2023-03-06
Add `required` to the CLI arguments used to spin up the `docker compose` instance

### LINKED ISSUE #75: Make the required CLI arguments  use `Required`
CLI arguments/ env variables that are compulsory should be forced to be present with the click option `Required=True`.

This will stop a bunch of `None` being passed to the PyDantic models. May help with #50 but semi orthogonal.

hmacdope:
Fixes #75 

As far as I can tell I have added required to the CLI arguments that are compulsory for spinning up the instance. I couldn't verify that `docker compose` worked as designed as I don't have the `aws` and `cloudwatch` keys. Perhaps @mikemhenry can confirm? 

Other than that should be good to go.

### DIFF
--- a/alchemiscale/cli.py
+++ b/alchemiscale/cli.py
@@ -312,7 +312,7 @@ def synchronous():
     ...
 
 
-@cli.group()
+@cli.group(help="Subcommands for the database")
 def database():
     ...
 
@@ -412,7 +412,7 @@ def scope(func):
     return scope(func)
 
 
-@cli.group()
+@cli.group(help="Subcommands for managing identities")
 def identity():
     ...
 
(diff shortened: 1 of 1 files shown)


mikemhenry:
I will test this, but this is also a good excuse for me to add add some docs to the README about how to spin up a test instance of our stack locally.

mikemhenry:
@hmacdope I haven't tested this but does `required=True` respect if there is an envar set?

mikemhenry (approved):
:tada: Looks good to me! I haven't gotten around to testing it yet but since it can accept envars, then I know it works since that how we currently do it :)

PR #103, opened 2023-03-20
Add `set_tasks_status` and `get_tasks_status` method to AlchemiscaleClient

### LINKED ISSUE #102: Add `set_tasks_status` method to `AlchemiscaleClient`; only allow setting to `waiting`, `invalid`, `deleted` via API
We want users to be able to set a task's status to any one of `waiting`, `invalid`, or `deleted`, as permitted by the storage layer. We need a `set_tasks_status` method on the `AlchemiscaleClient` that does this, with `AlchemiscaleAPI` enforcement that only these statuses are allowed to be set to by a user.

The corresponding `Neo4jStore` methods already enforce what statuses are allowed to transition to `waiting`, `invalid`, or `deleted`, so there will be no need to do any additional enforcement.

hmacdope:
Fixes #102

- Added task setters
- Made it return a list and made methods for setting a single task to complete private.
- Corresponding changes in testsuite
- Also made matching changes in `ComputeClient`


hmacdope:
Ready for a first look @dotsdl

dotsdl (approved):
Looks great @hmacdope! I removed the `set_tasks_status` and `get_tasks_status` methods from the compute client, and corresponding methods in the compute API, since we don't technically need them compute-side.

Merge when satisfied!

PR #104, opened 2023-03-24
Result path conversion and upload to object store

dotsdl:
Still a WIP; separating this work out from #98.


dotsdl:
We want to align this with the work being performed in OpenFreeEnergy/gufe#186 and OpenFreeEnergy/gufe#234. To accomplish this, I've articulated a proposal that subsumes the work in this PR in #180.

PR #113, opened 2023-04-04
Update Deployment Docs

mikemhenry:
* Update deployment docs to reflect using a launch template
* Update to Amazon Linux 2023
* Add HOSTNAME to log path so we can tell which instance the logs are coming from  
* Add tip on how to make JWT secret


hmacdope:
@mikemhenry am I okay to try this out on our AWS? (will there be an issue with me spinning up another instance?)

That way I can see if I can follow along and give better feedback.

docker/README.md
```diff
@@ -27,19 +48,49 @@ mkdir -p data/server
...
+
+# Annex
+
+The stuff below is not needed for deployment but useful to reference.
+
+## AWS EC2 Setup
+
+The following series of commands are sufficient for setting up Docker on an AWS EC2 instance running Amazon Linux. 
+Updated for Amazon Linux 2023.
+However, please use the launch template XXXX to spin up a new instance as it will use an AMI that has all dependences installed (so you will not have to do the steps below).
```
hmacdope, line 89: is XXXX to be filled in now or later?

mikemhenry: Lmao, I wrote that before I made the launch template.

mikemhenry:
@hmacdope great idea! See if you can setup `api.dev.alchemiscale.org` and `compute.dev.alchemiscale.org` on your own

docker/README.md
```diff
@@ -1,9 +1,27 @@
+# Launch New Instance
+
+1. Use "Launch instance from template"
+Use the `alchemiscale-template` to lauch a new EC2 instance.
+The only setting that will need modification is under "Key pair (login)".
+Select "Proceed without a key pair" (SSH keys have already be added to the image).
+
+2. Setup DNS
+Copy the public IP from the newly launched instance.
+Use that IP address to create an A record with desired hostname.
```
hmacdope, line 10: Would be useful to have some instructions on how to do this for a rank AWS beginner like me.

docker/README.md
```diff
@@ -27,19 +48,49 @@ mkdir -p data/server
...
 4. Start the service
- 
+
 ```bash 
 USER_ID=$(id -u) GROUP_ID=$(id -g) docker-compose up -d
 ```
 
-# AWS EC2 Setup
+Congratulations! 
+Check to see if everything is working by visiting `HOSTNAME/ping`
```
hmacdope, line 57: Could you give a quick example?

hmacdope (changes requested):
Two super quick comments.

docker/README.md
dotsdl: These instructions assume a user deploying `alchemiscale` to their AWS environment has an `alchemiscale-template` defined. This is not generally true, and the `docker-compose`-based instructions are a better fit for generality.

I think instructions like these may make sense for an organization's internal docs on how to re-deploy a host from an existing template, but they don't belong in the `alchemiscale` codebase repo itself.

docker/README.md
```diff
@@ -15,6 +33,9 @@ See `.env.testing` for an example.
 `- "--certificatesresolvers.myresolver.acme.caserver=https://acme-staging-v02.api.letsencrypt.org/directory"`
 in the docker-compose.yml file if you want Let's Encrypt (LE) to issue a real SSL certificate.
 The API for LE is rate limited, so I recommend first keeping this line in to see if a certificate is issued.
+Then you can comment out this line.
+You are allowed 5 failures per account, per hostname, per hour so it is better to make sure the staging cert is issued.
+It will show up as unsecured in a web browser but it will be signed by Let's Encrypt.
```
dotsdl, line 38: Thanks for this!

docker/README.md
dotsdl: Let's be more specific: it should be `https://api.${HOST_DOMAIN}/ping` for the user-facing API. This will only work, however, if an A record exists for the `api` subdomain, which we don't have explicitly in our instructions.

docker/README.md
```diff
@@ -27,19 +48,49 @@ mkdir -p data/server
...
+
+# Annex
+
+The stuff below is not needed for deployment but useful to reference.
+
+## AWS EC2 Setup
+
+The following series of commands are sufficient for setting up Docker on an AWS EC2 instance running Amazon Linux. 
+Updated for Amazon Linux 2023.
+However, please use the launch `template alchemiscale-template` to spin up a new instance as it will use an AMI that has all dependences installed (so you will not have to do the steps below).
```
dotsdl, line 89: From my comment above: these are not unnecessary, they are crucial. These should be our canonical instructions for deploying `alchemiscale` to an AWS environment, ideally with no assumptions on pre-existing resources defined.

docker/.env.template
```diff
@@ -15,6 +15,8 @@ AWS_SECRET_ACCESS_KEY=
 AWS_SESSION_TOKEN=
 
 # JWT
+# Tip to make JWT Secret
+# python -c "import secrets; print(secrets.token_hex(32))"
```
dotsdl, line 19: Ah yes, this is a good place for this!

dotsdl (changes requested):
@mikemhenry thanks for putting this together! Although I understand that using a launch template makes things easier for *repeated* deployments, I don't think the framing here is appropriate for the deployment docs in the `alchemiscale` repo itself.

We want these docs to be general enough for *any* potential user with an AWS account to use to get up and running; they should not be written assuming a specific AWS environment is being used with pre-defined resources.

docker/README.md
```diff
@@ -1,9 +1,27 @@
...
+The username is `ec2-user`.
+For example: `ssh ec2-user@compute.alchemiscale.org`
+
 # Deployment via `docker-compose`
 
 Note: this assumes a running Docker daemon and the `docker-compose` plugin.
 
 To deploy (using bash + EC2 host):
 
+0. `cd` into the correct directory `cd /home/ec2-user/alchemiscale/docker`
```
hmacdope, line 23: Can't comment on the right line but I think it might also be worth mentioning that you need to add `HOST_DOMAIN`  equal to whatever host domain you set in the A records and `ALCHEMISCALE_DOCKER_IMAGE` to be a recent image off `ghcr.io` the default one in `.env.testing` is an old one off a feature branch.

mikemhenry:
Okay! I understand now. Stuff that is specific to our AWS deployment I will migrate to a notion page so that we have the internal documentation, and stuff that along the lines of "how to deploy this stack on an arbitrary host" will live in the docs for this package. In my next pass at this I will migrate the org specific stuff to a notion page and move the general stuff somewhere like  `docs/deploy.rst` or something.

dotsdl:
Awesome, thank you @mikemhenry!

mikemhenry:
Superseded by #157

PR #115, opened 2023-04-04
Fix doc warnings

mikemhenry:
This fixes all the warnings in the docs, and also makes warnings show up as errors. 
This should then make it easier to stay on top of.


hmacdope:
@dotsdl I might go ahead and merge here if you are good with that?

dotsdl:
Good to merge once checks are green!

dotsdl (approved):
Thanks @mikemhenry, and sorry for the delay. This is good to merge!

PR #119, opened 2023-04-12
Separate deployment envs for server, compute

dotsdl:
Separated environments for server and compute deployment. Updated Docker build Action and Dockerfiles to support separate images for these.


dotsdl:
@mikemhenry any idea what's happening here: https://github.com/openforcefield/alchemiscale/actions/runs/4674041331/jobs/8277868408

Getting a failure on my attempt to build the images from this branch. It's not clear to me why this is happening.

hmacdope:
@dotsdl I think `dockerignore` might be excluding the env files:

```
# Ignore everything
*

# Only allow
!devtools/conda-envs/docker.yml <- needs to be explicitly listed.
!alchemiscale/*
!LICENSE 
!MANIFEST.in 
!README.md 
!setup.cfg 
!setup.py 
!versioneer.py

```

mikemhenry:
It is a dockerignore issue. I wanted to make sure we didn't accidentally included files we didn't want in the context. 

While I don't think that is a bad idea, since we are copying files into our image and not relying what is in the /docker folder, I don't think we need to be as strict.   

My suggestion is to follow @hmacdope advice here https://github.com/openforcefield/alchemiscale/pull/119#issuecomment-1504640933 and add `!devtools/conda-envs/docker.yml` and anything else you need to the dockerignore file. 

Once this is working, I would like to take a pass at it. I've recently learned a few new tricks from working on the openfe docker images. For example, we can build these images in a way that lets us share a docker daemon so that images are cached, which will speed up the build since they have the same base layers.

mikemhenry:
Also @dotsdl while you are working on this, I recomend setting the CI file to build docker images on this branch so you don't have to manually fire them off when you want to test - just a suggestion!

dotsdl:
I think this is ready for a review from you @mikemhenry! I'm going to do some live testing with the server and compute images in parallel, but feel free to push changes as you see fit.

mikemhenry:
Cool, I am busy with the openfe release so I won't be able to make some of the changes to speed up the build, but I can review this now for correctness

dotsdl:
Thanks @mikemhenry, no problem! We can optimize the build at a later time; it's plenty fast as-is.

mikemhenry (approved):
Just a few notes

PR #143, opened 2023-06-12
Add AlchemiscaleClient API points for simpler status, result retrieval

### LINKED ISSUE #126: [ENH] Ability to see summary of task status at org / campaign / project level
As far as I'm aware this isn't available using the client (at least not through the documentation provided to initial users).

It would be great if it would be possible to see the status of all tasks for an org / campaign / project using a single command rather than have to iterate through all the transformations. This would make it easier to track how things are going.

dotsdl:
Closes #126.


alchemiscale/storage/statestore.py
```diff
@@ -1570,6 +1698,56 @@ def get_task_transformation(
...
+
+        prop_string = ", ".join(
+            "{}: '{}'".format(key, value)
+            for key, value in properties.items()
+            if value is not None
+        )
+
+        q = f"""
+        MATCH (n:Task {{{prop_string}}})
+        RETURN n.status AS status, count(n) as counts
```
hmacdope, line 1718: Very nice.

alchemiscale/storage/statestore.py
```diff
@@ -675,29 +676,113 @@ def query_networks(
...
+            res = tx.run(q)
+            for rec in res:
+                sks.append(rec["sk"])
+
+        return [ScopedKey.from_str(sk) for sk in sks]
+
+    def get_network_chemicalsystems(self, network: ScopedKey) -> List[ScopedKey]:
+        """List ScopedKeys for ChemicalSystems associated with the given AlchemicalNetwork."""
+        q = f"""
+        MATCH (n:AlchemicalNetwork {{_scoped_key: '{network}'}})-[:DEPENDS_ON]->(cs:ChemicalSystem)
```
hmacdope, line 726: Same here,  but note the above dropping tags are just a  suggestion, I haven't tested the performance myself to see if it makes any measurable difference, if you want to keep them no worries. I'll stop making suggestions from here, I think you get the idea.

alchemiscale/tests/integration/interface/client/test_client.py
```diff
@@ -254,6 +639,18 @@ def test_cancel_tasks(
 
         assert canceled_sks_2 == [None]
 
+    def test_get_tasks_status(self):
+        ...
```
hmacdope, line 643: This and below to be finished this PR or another?

alchemiscale/tests/integration/storage/test_statestore.py
```diff
@@ -1019,6 +1093,15 @@ def test_get_task_transformation(
         assert tf == transformation
         assert tf_sk == transformation_sk
 
+    def test_get_scope_status(self, n4js: Neo4jStore):
```
hmacdope, line 1096: Same here, this PR or another one?

hmacdope (changes requested):
A few super minor changes @dotsdl. Other than that looks fantastic, this will make a huge difference to the user experience.

alchemiscale/interface/api.py
```diff
@@ -102,6 +103,25 @@ def check_existence(
...
+    n4js: Neo4jStore = Depends(get_n4js_depends),
+    token: TokenData = Depends(get_token_data_depends),
+):
+    validate_scopes(scope, token)
+
+    an = AlchemicalNetwork.from_dict(network)
+    an_sk = n4js.create_network(network=an, scope=scope)
+
+    # create taskhub for this network
+    n4js.create_taskhub(an_sk)
```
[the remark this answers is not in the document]
dotsdl: No, `TaskHub`s are an internal detail to the state store; since an `AlchemicalNetwork` has one and only one of these, we use the `AlchemicalNetwork`'s `ScopedKey` to perform operations, such as actioning tasks, on its `TaskHub` via the `AlchemiscaleClient`.

alchemiscale/interface/api.py
```diff
@@ -128,43 +147,140 @@ def query_networks(
...
+    validate_scopes(sk.scope, token)
+
+    return [
+        str(sk) for sk in n4js.get_transformation_chemicalsystems(transformation=sk)
+    ]
+
+
+@router.get("/chemicalsystems/{chemicalsystem_scoped_key}/transformations")
+def get_chemicalsystem_transformations(
+    chemicalsystem_scoped_key,
```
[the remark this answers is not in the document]
dotsdl: For the second case, would it give back a list of lists, corresponding to the order of `ScopedKey`s given in `chemsys_sks`? 

I suppose it could be a useful pattern to give back only a single item if a single item is given, an iterable of an iterable is given. For the iterable case, it may make sense for it to perform `async` calls for greater performance.

I'd rather expand these methods to do multi support if there is user demand. There are relatively few cases where I think there's high demand upfront (e.g. setting or getting many `Task` statuses), so we'll have to see how users are bottlenecked in practice. Agree?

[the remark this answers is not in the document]
hmacdope: Sounds good to me :)

alchemiscale/interface/api.py
```diff
@@ -233,19 +344,76 @@ def create_tasks(
...
     return_as: str = "list",
+    status: str = None,
     n4js: Neo4jStore = Depends(get_n4js_depends),
     token: TokenData = Depends(get_token_data_depends),
 ):
     sk = ScopedKey.from_str(transformation_scoped_key)
     validate_scopes(sk.scope, token)
 
-    task_sks = n4js.get_tasks(sk, extends=extends, return_as=return_as)
+    if status is not None:
```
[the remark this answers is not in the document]
dotsdl: :grin:

alchemiscale/models.py
```diff
@@ -24,6 +24,15 @@ def __str__(self):
         )
         return "-".join(triple)
 
+    def __lt__(self, other):
+        return str(self) < str(other)
+
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
```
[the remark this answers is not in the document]
dotsdl: It's just to avoid two different objects counting as equal if they come from different classes but resolve to the same string representation.

alchemiscale/storage/statestore.py
dotsdl: All good, appreciate pointing it out!

alchemiscale/tests/integration/interface/client/test_client.py
dotsdl: Another PR. I noticed that they were missing, but since we're going to be reworking the `status` methods and haven't implemented the `priority` methods yet, I added the placeholders.

alchemiscale/tests/integration/storage/test_statestore.py
dotsdl: I'll fill these in now. I prioritized getting the `AlchemiscaleClient` tests in for the API points, since these end up exercising the `Neo4jStore` methods anyway, though that will be invisible to codecov due to how they get called.

alchemiscale/storage/statestore.py
```diff
@@ -675,29 +676,113 @@ def query_networks(
...
         return self._query(
             qualname="ChemicalSystem", additional=additional, key=key, scope=scope
         )
 
-    def get_transformations_for_chemicalsystem(self):
-        ...
+    def get_network_transformations(self, network: ScopedKey) -> List[ScopedKey]:
+        """List ScopedKeys for Transformations associated with the given AlchemicalNetwork."""
+        q = f"""
+        MATCH (n:AlchemicalNetwork {{_scoped_key: '{network}'}})-[:DEPENDS_ON]->(t:Transformation)
```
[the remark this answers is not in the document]
dotsdl: Got it!

dotsdl:
Just finished second round of changes in response to your review @hmacdope! Can you give me an approve if you're satisfied?

hmacdope (approved):
Looks great @dotsdl! Thanks for the hard work. Merge when ready if CI is happy.

PR #155, opened 2023-06-30
`AlchemiscaleClient` LRU caching

dotsdl:
Added LRU caching to expensive `AlchemiscaleClient` methods.

We add LRU caching to the following methods:
- `get_network`
- `get_transformation`
- `get_chemicalsystem`

and to the underlying `ProtocolDAGResult` retrieval method for:
- `get_transformation_results`
- `get_transformation_failures`
- `get_task_results`
- `get_task_failures`

This is just an optimization; it does not impact the results of these method calls. Since their target objects are immutable, it is safe to cache these client-side.

We also add a `visualize` kwarg to each of these methods that allows the user to switch off the `rich`-based progress visualizations.

### DIFF
--- a/alchemiscale/interface/client.py
+++ b/alchemiscale/interface/client.py
@@ -220,7 +220,10 @@ class AlchemiscaleClient(AlchemiscaleBaseClient):
 
     @lru_cache(maxsize=100)
     def get_network(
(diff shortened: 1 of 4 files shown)


hmacdope (approved):
Look great, thanks @dotsdl

PR #158, opened 2023-07-07
Update prod envs to use openfe 0.11.0

dotsdl:


dotsdl:
Tested on QA host for `alchemiscale.org`. Users will need `openfe=0.11.0` client-side in order to handle new `output_structure` field in `openfe.protocols.openfe_rfe.equil_rfe_settings.SimulationSettings`. This field is optional, so old `SimulationSettings` instances in an existing DB will deserialize fine.

PR #160, opened 2023-08-06
Set pydantic version less than 2.0 for all conda envs

dotsdl:
Pydantic 2.0 comes with major changes that we are not yet ready to support, and installs of it in environments that use `alchemiscale` are likely to break usage. This PR sets the version spec to <2.0 for pydantic in all conda envs.

### DIFF
--- a/alchemiscale/compute/api.py
+++ b/alchemiscale/compute/api.py
@@ -204,16 +204,30 @@ def get_task_transformation(
     sk = ScopedKey.from_str(task_scoped_key)
     validate_scopes(sk.scope, token)
 
-    transformation, protocoldagresultref = n4js.get_task_transformation(
-        task=task_scoped_key
+    transformation_sk, protocoldagresultref_sk = n4js.get_task_transformation(
+        task=task_scoped_key, return_gufe=False
     )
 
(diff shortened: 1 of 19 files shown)


hmacdope:
Just FYI if you want to support pydantic~1 schema from pydantic>=2.0.0a @mattwthompson, @jhorton and I have been using this to import the v1 api from v2. 

```python
    from pydantic.v1 import BaseModel, Field
```

You can then use the following guard to support both versions.

```python
try:
    from pydantic.v1 import BaseModel, Field
except ImportError:
    from pydantic import BaseModel, Field
```

dotsdl:
Oh gotcha, I suppose that would be the preferred approach to avoid holding back the rest of the stack?

dotsdl:
Trying out the import guard approach; will see if this breaks our tests.

dotsdl:
Reverting to just supporting `pydantic < 2.0` for now. We'll work toward proper `pydantic` 2.0 support in a future PR; tests gave an opaque error when we tried doing the import guard approach above: https://github.com/openforcefield/alchemiscale/actions/runs/5791301527/job/15695900360?pr=160

PR #161, opened 2023-08-06
Make `ProtocolDAGResult` push and pull keys self-consistent

dotsdl:
This PR fixes opportunities for mismatch between the `gufe` keys of `Transformation`s compute-side vs. state store-side. This mismatch impacted `ProtocolDAGResult`s generated and pushed compute-side, ultimately leading results that were inaccessible to users client-side. In other words, this led to inconsistency between the state store and object store.

There are two components to this PR:
- fix to compute-side pushes such that the `Transformation` `ScopedKey` *as given by the state store* is also used for `ProtocolDAGResult` pushes; for new `ProtocolDAGResult`s, this should be sufficient to avoid mismatches entirely
- addition of fallback retrieval based on `ProtocolDAGResultRef` `location` when retrieval via object store path construction fails; this should address existing mismatches in an `alchemiscale` deployment


alchemiscale/interface/api.py
```diff
@@ -640,10 +640,7 @@ def get_transformation_results(
     sk = ScopedKey.from_str(transformation_scoped_key)
     validate_scopes(sk.scope, token)
 
-    # get all ProtocolDAGResultRefs for the given transformation's results
-    refs: List[ProtocolDAGResultRef] = n4js.get_transformation_results(sk)
-
-    return [i.to_dict() for i in refs]
+    return [str(sk) for sk in n4js.get_transformation_results(sk)]
```
hmacdope, line 643: just a query, because I remember discussing this at some point, is the pattern to return `str(sk)` or `sk` objects in general.

alchemiscale/storage/objectstore.py
```diff
@@ -264,22 +274,23 @@ def pull_protocoldagresult(
...
-            "protocoldagresult",
-            *protocoldagresult.scope.to_tuple(),
-            transformation.gufe_key,
-            route,
-            protocoldagresult.gufe_key,
-            "obj.json",
-        )
+        # build `location` based on provided ScopedKey if not provided
+        if location is None:
+            if transformation.scope != protocoldagresult.scope:
```
hmacdope, line 281: Should also check that `transformation` and `protocoldagresult` are not both None as is default. Ie can get this situation 
```python
location == None 
transformation == None 
protocoldagresult == None
```
Which will have a fit when trying to check 

```python
if transformation.scope != protocoldagresult.scope:
```

alchemiscale/storage/statestore.py
```diff
@@ -818,7 +816,8 @@ def get_transformation_failures(
         MATCH (trans:Transformation {_scoped_key: $scoped_key}),
               (trans)<-[:PERFORMS]-(:Task)-[:RESULTS_IN]->(res:ProtocolDAGResultRef)
         WHERE res.ok = false
-        RETURN res
+        WITH res._scoped_key as sk
+        RETURN DISTINCT sk
```
hmacdope, line 820: Query but under what situations would it be possible to end up with duplicate sks here?

hmacdope (approved):
Looks great @dotsdl, important fix. Few queries and one suggestion

alchemiscale/storage/statestore.py
dotsdl: We end up with them in the tests, where we push the same `ProtocolDAGResultRef` for two different `Task`s on the same `Transformation`. Not really a case we should see in real usage, but forcing `DISTINCT` here avoids the issue. We really do only want one instance of a given `ProtocolDAGResult` coming out of this method.

alchemiscale/interface/api.py
dotsdl: This is for simpler JSON-serializability of `ScopedKey`s, since they can be fully-represented as strings. We could leave them as-is since they're just `pydantic` models, but they'd just have to processed as dicts instead client-side. Ends up being marginally easier this way IMHO.

alchemiscale/storage/objectstore.py
dotsdl: Ah good call; adding in a guardrail here.

PR #162, opened 2023-08-08
Only action `Task`s that are `status` 'waiting', 'error', or 'running'

### LINKED ISSUE #138: Make it impossible to action `Task`s that don't have `status` 'waiting' or 'error'
It looks like it's currently possible to action `Task`s that are any `status`. We don't want this to be possible, as only `Task`s that can conceivably be run yet (`status` is 'waiting' or 'error'). It shouldn't impact downstream compute, but does create unnecessary ACTIONS relationships to a `TaskHub` that will likely never be cleared.

We should add a `status` check in the Cypher query for `Task`s actioned in `Neo4jStore.action_tasks`. We will return `None` for any `Task` that was requested to be actioned but ultimately was not; this should be the resulting behavior with this change.

dotsdl:
Closes #138.

We also include `'running'` as a valid `status` for actioning `Task`s, as a currently `'running'` `Task` may end up returning to either `'waiting'` or becoming status `'error`'. In other words, a user may reasonably want to action any non-complete/deleted/invalid `Task` on a given `AlchemicalNetwork` featuring that `Task`'s `Transformation`.

### DIFF
--- a/alchemiscale/interface/client.py
+++ b/alchemiscale/interface/client.py
@@ -607,6 +607,7 @@ class AlchemiscaleClient(AlchemiscaleBaseClient):
         A Task cannot be actioned:
             - to an AlchemicalNetwork in a different Scope.
             - if it extends another Task that is not complete.
+            - if it has any status other than 'waiting', 'running', or 'error'
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


hmacdope (approved):
LGTM!

PR #170, opened 2023-08-23
build and test all images, add apptainer

mikemhenry:


mikemhenry:
I was shocked when this was all green, but I forgot to set it up to actually run the docker building :face_exhaling:

PR #172, opened 2023-08-24
Support adding and removing multiple scopes from identity with single CLI call

dotsdl:


alchemiscale/tests/integration/test_cli.py
mikemhenry, line 1: Just one thought, what if we made this a parameterized test, where we test adding a single scope and also more than one, so that we make sure we don't make a regression if someone only wants to add one?

dotsdl: Done! Thanks for this suggestion @mikemhenry!

PR #174, opened 2023-08-27
Add Alchemiscale Demo as first tutorial

dotsdl:
This demo is based on a notebook originally created by @ialibay. 
It illustrates:

1. Creating an AlchemicalNetwork
2. Submitting an Alchemicalnetwork to alchemiscale
3. Monitoring the progress of an alchemiscale submission
4. Gathering results and analyzing them using cinnabar

This PR also adds use of the MystNB Sphinx extension to the doc configuration.


IAlibay (changes requested):
@dotsdl there's some changes to the notebook's plotting section I'd like to do, I was just about to push some changes before I went on leave at the end of last week - namely just using cinnabar's plotting directly (the colour scheme isn't amazing and the plots aren't flexible [yet] but at least we will avoid visual artifacts - like the fact that I failed to enforce square plots here).

I'll try to push some changes to the source notebook tomorrow if that's ok?

dotsdl:
@IAlibay of course! Let me know when you're happy with the current state of the original source notebook (here? https://github.com/IAlibay/alchemiscale-utilities/blob/main/notebooks/TYK2_MST/Alchemiscale_submit_and_analyze.ipynb) and I'll pull those bits in for the closing example!

dotsdl:
@mikemhenry since you recently dogfooded this content, if you've got time I'd value a review from you on how we've incorporated it here. :pray:

mikemhenry:
I was able to use this notebook and it worked without me having to make any modifications :tada: 

I think the only thing to consider if there are some operations that a user might not want to re-run, like this bit:

```
tasks = []
for tf_sk in asc.get_network_transformations(an_sk):
    tasks.extend(asc.create_tasks(tf_sk, count=1))

tasks
```

We might want to add logic so that we only make the tasks once, later on in the notebook we talk about how to extend, re-run, and set the status of tasks so we teach a user later how to do that, but before we get to that section, it would be nice to prevent a user from accidentally creating a ton of tasks.

IAlibay:
Sorry about the delay here, I've updated the notebook - just waiting on results to come back so we get proper plots.

dotsdl:
@IAlibay all good! I'll take what you've put together there and incorporate into the tutorial here!

dotsdl:
> We might want to add logic so that we only make the tasks once, later on in the notebook we talk about how to extend, re-run, and set the status of tasks so we teach a user later how to do that, but before we get to that section, it would be nice to prevent a user from accidentally creating a ton of tasks.

@mikemhenry we do show something like this in the [User Guide](https://docs.alchemiscale.org/en/latest/user_guide.html#creating-and-actioning-tasks), but I think I'd like to keep the tutorial as simple as possible here. I would like to finally get a `set_transformation_tasks` method working, which would idempotently set the number of tasks to the given count if it isn't already there, but we don't have it yet.

@IAlibay I've added your changes; looking great! Did I miss anything? https://alchemiscale--174.org.readthedocs.build/en/174/tutorials/demo/Alchemiscale%20Demo.html

docs/tutorials/demo/Alchemiscale Demo.ipynb
```diff
@@ -0,0 +1,2774 @@
...
+   "source": [
+    "We first illustrate how to build `AlchemicalNetwork` objects,\n",
+    "from a starting point of chemical models stored in sdf and pdb files.\n",
+    "\n",
+    "An `AlchemicalNetwork` is used to represent an entire network of calculations,\n",
+    "and is composed of many smaller objects:\n",
+    "\n",
+    "- An `AlchemicalNetwork` composed of \n",
+    "  - each node a `ChemicalSystem`\n",
+    "    - each containing many components, such as `SmallMoleculeComponent`, `ProteinComponent`\n",
```
IAlibay, line 50: Would it make sense to have these link to the original docs? https://docs.openfree.energy/en/stable/reference/api/systems_and_components.html etc..

docs/tutorials/demo/Alchemiscale Demo.ipynb
```diff
@@ -0,0 +1,2774 @@
...
+       " 'system_settings': {'nonbonded_method': 'PME',\n",
+       "  'nonbonded_cutoff': 1.0 <Unit('nanometer')>},\n",
+       " 'solvation_settings': {'solvent_model': 'tip3p',\n",
+       "  'solvent_padding': 1.2 <Unit('nanometer')>},\n",
+       " 'alchemical_settings': {'lambda_functions': 'default',\n",
+       "  'lambda_windows': 11,\n",
+       "  'unsampled_endstates': False,\n",
+       "  'use_dispersion_correction': False,\n",
+       "  'softcore_LJ_v2': True,\n",
+       "  'softcore_electrostatics': True,\n",
```
IAlibay, line 435: I believe the softcore_electrostatics entry have been removed now as of ~ openfe 0.12?

docs/tutorials/demo/Alchemiscale Demo.ipynb
```diff
@@ -0,0 +1,2774 @@
...
+   "id": "c38305b0-879f-43b3-ba13-b010e478d2eb",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Here we make it so that each simulation only encompasses a single repeat\n",
+    "# We then do multiple repeats by running each simulation multiple time\n",
+    "protocol_settings.alchemical_sampler_settings.n_repeats = 1\n",
+    "\n",
+    "# We enforce the compute platform to be CUDA. This ensures that a bad GPU node\n",
+    "# on alchemiscale will fail automatically rather than trying to defaul to the CPU kernels\n",
```
IAlibay, line 490:
```suggestion
    "# on alchemiscale will fail automatically rather than trying to default to the CPU kernels\n",
```

docs/tutorials/demo/Alchemiscale Demo.ipynb
```diff
@@ -0,0 +1,2774 @@
...
+    "protocol_settings.engine_settings.compute_platform = \"CUDA\"\n",
+    "\n",
+    "# We set the protocol to auto terminate once the error drops below 0.2 kT\n",
```
IAlibay, line 493:
```suggestion
    "# We set the protocol to auto terminate once the MBAR error of the estimate drops below 0.2 kT\n",
```

docs/tutorials/demo/Alchemiscale Demo.ipynb
```diff
@@ -0,0 +1,2774 @@
...
+     "execution_count": 21,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "network = AlchemicalNetwork(edges=(solvated_transformations + complexed_transformations), \n",
+    "                            nodes=(list(solvated.values()) + list(complexed.values())),\n",
+    "                            name=\"tyk2_relative_benchmark\")\n",
+    "network"
```
IAlibay, line 709: We've improved the network representtation post openfe 0.11, it should look a lot less cluttered

docs/tutorials/demo/Alchemiscale Demo.ipynb
```diff
@@ -0,0 +1,2774 @@
...
+   "source": [
+    "results['Transformation-08f4c68a8e178d698e59cc1a366e3415-ddotson-tyk2-demo'].get_uncertainty()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "60d404e8-11ab-4ead-b0db-eddaca9e2c51",
+   "metadata": {},
+   "source": [
+    "In this case, we have only a single `ProtocolDAGResult` for each `Transformation` (since we created and actioned only 1 `Task` for each), and so the uncertainty given for this `ProtocolResult` is 0.0. The `RelativeHybridTopologyProtocol` combines `ProtocolDAGResult` values statistically, reducing the uncertainty but not increasing convergence with additional `Task`s\n",
```
IAlibay, line 1798:
```suggestion
    "In this case, we have only a single `ProtocolDAGResult` for each `Transformation` (since we created and actioned only 1 `Task` for each), and so the uncertainty (standard deviation between replicate simulations) given for this `ProtocolResult` is 0.0. The `RelativeHybridTopologyProtocol` combines `ProtocolDAGResult` values statistically, reducing the uncertainty but not increasing convergence with additional `Task`s\n",
```

docs/tutorials/demo/Alchemiscale Demo.ipynb
```diff
@@ -0,0 +1,2774 @@
...
+      "    type: ki\n",
+      "    unit: nM\n",
+      "    value: 8.5\n",
+      "  name: lig_jmc_28\n",
+      "  smiles: '[H]c1c(c(c(c(c1[H])Cl)C(=O)N([H])c2c(c(nc(c2[H])N([H])C(=O)[C@@]3([C@](C3([H])[H])([H])C([H])([H])[H])[H])[H])[H])Cl)[H]'\n"
+     ]
+    }
+   ],
+   "source": [
+    "%cat ligands.yml"
```
IAlibay, line 2380: You don't really explain what's in this file here? Probably something like "we have a PLB yaml file with experimental data for each ligand in our network"

IAlibay (commented):
Just a couple of small things.

docs/tutorials/demo/Alchemiscale Demo.ipynb
dotsdl: Fair enough; I think I should show the versions of `openfe`, `gufe`, and `alchemiscale` used for the notebook generation. I'll add these.

docs/tutorials/demo/Alchemiscale Demo.ipynb
dotsdl: How do you mean? This is just defining the network itself; do you mean some kind of visualization?

docs/tutorials/demo/Alchemiscale Demo.ipynb
dotsdl: Added clarification!

docs/tutorials/demo/Alchemiscale Demo.ipynb
dotsdl: Not sure how stable those specific page links will be, so would rather not here. We do link to the `openfe` repo at the top of the notebook.

docs/tutorials/demo/Alchemiscale Demo.ipynb
dotsdl: Done!

dotsdl:
Addressed your comments @IAlibay! Thanks for these!

PR #177, opened 2023-09-02
Review and refinement of API docs

### LINKED ISSUE #32: Docs : API reference
We need API reference docs for the components referenced in the other docs (#28, #29, #30, #31). These are rather easy to generate with Sphinx, but closing this issue will require adding or improving docstrings where they are clearly missing or insufficient in the public methods/functions in the codebase.

dotsdl:
Working on making the API docs more polished.
Closes #32.

### DIFF
--- a/alchemiscale/base/api.py
+++ b/alchemiscale/base/api.py
@@ -1,9 +1,8 @@
 """
-Reusable components for API services. --- :mod:`alchemiscale.base.api`
-======================================================================
+:mod:`alchemiscale.base.api` --- reusable components for API services
+=====================================================================
 
 """
-
 
 from functools import lru_cache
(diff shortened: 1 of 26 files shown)


dotsdl:
I think I've gone about as far as I'd like to go on this. There are surely objects and methods that could use better documentation, but this is better done gradually as we notice gaps than trying to do it in one PR.

hmacdope (approved):
Looks good to me @dotsdl. One thing I might add is mentioning the permissive MIT licensing in developer docs, or even on home page of the docs (only if you like).

dotsdl:
Thanks @hmacdope! Added this to the top-level index for the docs!

PR #178, opened 2023-09-02
Added `get_network_results` method to `AlchemiscaleClient`

### LINKED ISSUE #154: Add `get_network_results` method to `AlchemiscaleClient`
Currently, the highest-level way of getting results back exposed in the `AlchemiscaleClient` is via `get_transformation_results`, which operates at the level of a single `Transformation`. A typical user will likely loop through all `Transformation`s in their `AlchemicalNetwork` of interest one at a time, and though this method now benefits from `async` retrieval of multiple `ProtocolDAGResult`s at a time, this usage pattern isn't exactly optimal.

We should add a `get_network_results` method that returns a dict giving the `ScopedKey` of each `Transformation` as keys, `ProtocolResult` (or list of `ProtocolDAGResult`s, similar to `get_transformation_results`) as values. This can be pretty fast given that all retrievals can be done in a single batch of `async` calls, and with LRU caching of objects can be run repeatedly with high performance.

dotsdl:
Closes #154.

There are a couple ways we could go about this. It's unclear which one
will yield higher performance. Will try both and compare:
1. retrieve results for each `Transformation` as a separate task for a
   process pool, with each using `asyncio` separately
2. toss all `ProtocolDAGResult` requests into the same `asyncio` event loop

(1) may make better use of available CPU resources, since a bottleneck
when retrieving many or large `ProtocolDAGResult`s is deserialization
from JSON. So that deserialization can be done over many processes, but
an additional pickle serialization/deserialization is required to return
these objects to the parent process.

(2) would be more efficient if the bottleneck really is just network
latency, not deserialization.

We may end up providing both (1) and (2), and make them switchable via a
kwarg. This will allow users to choose the approach that yields better
performance for them, which can be geography and local network
speed-dependent.

### DIFF
--- a/alchemiscale/base/api.py
+++ b/alchemiscale/base/api.py
@@ -255,4 +255,8 @@ def get_access_token(
         jwt_algorithm=settings.JWT_ALGORITHM,
     )
 
-    return {"access_token": access_token, "token_type": "bearer", 'expiration': expire.isoformat()}
+    return {
+        "access_token": access_token,
+        "token_type": "bearer",
+        "expiration": expire.isoformat(),
+    }
--- a/alchemiscale/base/api.py
+++ b/alchemiscale/base/api.py
@@ -1,9 +1,8 @@
 """
-Reusable components for API services. --- :mod:`alchemiscale.base.api`
-======================================================================
+:mod:`alchemiscale.base.api` --- reusable components for API services
+=====================================================================
 
 """
-
 
 from functools import lru_cache
 from typing import Any, Union, Dict, List, Callable
--- a/alchemiscale/base/api.py
+++ b/alchemiscale/base/api.py
@@ -248,11 +248,11 @@ def get_access_token(
             headers={"WWW-Authenticate": "Bearer"},
         )
 
-    access_token = create_access_token(
+    access_token, expire = create_access_token(
         data={"sub": entity.identifier, "scopes": entity.scopes},
         secret_key=settings.JWT_SECRET_KEY,
         expires_seconds=settings.JWT_EXPIRE_SECONDS,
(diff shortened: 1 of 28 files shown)


dotsdl:
Using implementation (1) as:
```python
results = asc.get_network_results(an_sk)
```
on a test `tyk2` network takes about 15s, compared to 1min20s for the current recommended approach:
```python
results_ = {}
for tf_sk in asc.get_network_transformations(an_sk):
    results_[tf_sk] = asc.get_transformation_results(tf_sk, visualize=False)
```

So, looking good so far! Taking advantage of multiple processes to pull `Transformation`s and `ProtocolDAGResult`s, `gather` them, then just serialize/deserialize via pickle the `ProtocolResult` object appears to yield much better performance.

Seeing an pathology in JupyterLab where running the first block above *after* the second block causes the first block to hang indefinitely. Unclear if we're getting some weird interaction between Jupyter's use of an event loop and our use of `ProcessPoolExecutor`. Will investigate further.

dotsdl:
Chose to go with option (1), and got it working by forcing use of `'spawn`' as the multiprocessing context, as [what we are doing to get `async` methods working under Jupyter interferes with use of `'fork'`](https://github.com/erdewit/nest_asyncio/issues/81). This gives reasonable performance, between 5x and 10x depending on the number and size of `ProtocolDAGResult`s for each `Transformation` in an `AlchemicalNetwork` compared to using `AlchemiscaleClient.get_transformation_results` in a loop.

Ready for review! @mikemhenry you able to hit this one?

mikemhenry (approved):

mikemhenry:
Ready to merge at your leisure

dotsdl:
Awesome, thanks so much @mikemhenry!

PR #183, opened 2023-09-26
[doc] docker compose plugin to newest version

mikemhenry:
Testing newest compose plugin on a new deployment, if this works I will just merge it in, no review needed.


mikemhenry:
So far so good, but I want to make sure it works under load before we update the doc.

PR #193, opened 2023-10-19
Set minimum for openmmforcefields version in conda envs to 0.12.0

dotsdl:
Even though 0.12.0 is currently the latest version of `openmmforcefields`, building the env without forcing it appears to get 0.11.2 at the moment.

### DIFF
--- a/.github/workflows/ci-integration.yml
+++ b/.github/workflows/ci-integration.yml
@@ -29,6 +29,7 @@ jobs:
         python-version:
           - "3.9"
           - "3.10"
+          - "3.11"
     env:
       OE_LICENSE: ${{ github.workspace }}/oe_license.txt
 
--- a/.github/workflows/ci-integration.yml
+++ b/.github/workflows/ci-integration.yml
@@ -29,7 +29,6 @@ jobs:
         python-version:
           - "3.9"
           - "3.10"
-          - "3.11"
     env:
       OE_LICENSE: ${{ github.workspace }}/oe_license.txt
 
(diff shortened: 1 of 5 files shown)


dotsdl:
Looks like [ambertools is keeping us from using python 3.11](https://github.com/openforcefield/alchemiscale/actions/runs/6569599669/job/17845733732?pr=193#step:3:714)? @mikemhenry any insights as to why this may suddenly be the case with `openmmtools` 0.12.0?

I'm fine with downgrading to Python 3.10 in this instance, but wanted to raise this as odd in case it's the first we've seen it.

mikemhenry:
We don't have python 3.11 builds for older ambertools, and we can't use the newest openmmtools with the older ambertools. 

Does that make sense? In a perfect world, I would do a amber22 build for python 3.11

dotsdl:
@mikemhenry yes, that makes sense. As discussed today, can you take a stab at getting an ambertools 22 build for Python 3.11, since I think without that, practically the whole `openforcefield` stack and everything using it will be held back to Python 3.10. Not exactly urgent, but seems unlikely to resolve on its own?

For (my future) reference, here is the pin that holds us back from `ambertools` 23: https://github.com/conda-forge/openmmforcefields-feedstock/blob/543afe9bd8fc8e6dbc3084ba59d0bfeaa26a358b/recipe/meta.yaml#L25

dotsdl:
For the moment then here, we will remove Python 3.11 as a tested version, and proceed with deploying on Python 3.10. The CI for that version will always fail at present given the problem above.

dotsdl:
Decided to keep 3.11 CI test; we don't require it to pass before merge, so this keeps it around as a reminder that this needs fixing. May revisit this decision if it proves annoying having it around.

PR #203, opened 2023-11-20
Make test env use latest release of openfe and gufe instead of `main`

dotsdl:
Now that `openfe` and `gufe` are operating at regular release cadences, I think it makes sense for us to perform testing against the latest releases of these on `conda-forge` instead of testing against the `main` branches of them. In particular, this saves us the need of having to manually track their dependencies, such as the recently added `kartograf` dependency in `openfe`.

I think this is generally a safe move, since we only ever deploy production envs with releases of these packages too (pinned to specific releases). The CI/test env still operates as a canary of upstream changes, but on a delay dependent on upstream releases compared to current state.

### DIFF
--- a/.github/workflows/ci-integration.yml
+++ b/.github/workflows/ci-integration.yml
@@ -29,8 +29,8 @@ jobs:
         python-version:
           - "3.9"
           - "3.10"
-    env:
-      OE_LICENSE: ${{ github.workspace }}/oe_license.txt
+    #env:
+    #  OE_LICENSE: ${{ github.workspace }}/oe_license.txt
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


dotsdl:
@mikemhenry thoughts on this? Any reasons why we might not want to go this route?

devtools/conda-envs/test.yml
```diff
@@ -2,22 +2,15 @@ name: alchemiscale-test
...
 dependencies:
   - pip
 
-  # gufe dependencies
-  - numpy<1.24  # https://github.com/numba/numba/issues/8615#issuecomment-1360792615
-  - networkx
-  - rdkit
+  # alchemiscale dependencies
+  - gufe
+  - openfe
```
mikemhenry, line 10: not blocking but doing `>=` for min versions we support would save us from pulling in something old by mistake

mikemhenry (approved):

.github/workflows/ci-integration.yml
```diff
@@ -44,13 +44,13 @@ jobs:
...
-        run: |
-          echo "${OE_LICENSE_TEXT}" > ${OE_LICENSE}
-          python -c "import openeye; assert openeye.oechem.OEChemIsLicensed(), 'OpenEye license checks failed!'"
+      #- name: Decrypt OpenEye license
+      #  if: ${{ !github.event.pull_request.head.repo.fork }}
+      #  env:
+      #    OE_LICENSE_TEXT: ${{ secrets.OE_LICENSE }}
+      #  run: |
+      #    echo "${OE_LICENSE_TEXT}" > ${OE_LICENSE}
+      #    python -c "import openeye; assert openeye.oechem.OEChemIsLicensed(), 'OpenEye license checks failed!'"
```
mikemhenry, line 53:
```suggestion
```

.github/workflows/ci-integration.yml
```diff
@@ -29,8 +29,8 @@ jobs:
         python-version:
           - "3.9"
           - "3.10"
-    env:
-      OE_LICENSE: ${{ github.workspace }}/oe_license.txt
+    #env:
+    #  OE_LICENSE: ${{ github.workspace }}/oe_license.txt
```
mikemhenry, line 33:
```suggestion
```

mikemhenry:
LGTM, only other thing would be remove the stuff commented out since it looks like it works without it

dotsdl:
Awesome, thanks @mikemhenry! Merging!

PR #206, opened 2023-12-01
Remove `return_gufe` kwarg from `AlchemiscaleClient.query_networks`

### LINKED ISSUE #142: Remove `return_gufe` kwarg from `AlchemiscaleClient`
Currently, it's possible to have the `AlchemiscaleClient.query_networks` method return a dict of `AlchemicalNetwork`s, with `ScopedKey`s as keys and `AlchemicalNetwork`s as values. In practice, this is unlikely to get used, as it is unclear from a given query how many networks will come back, and so performance will almost always remain an issue.

Instead of this, we should simplify this method and the calls down through the `Neo4jStore` layer, removing the `return_gufe` functionality.

Users are instead encouraged to use the `query_networks` method for getting all `ScopedKey`s of `AlchemicalNetwork`s that match their query parameters, then use these with `get_network` to actually retrieve the networks of interest.

ianmkenney:
Fixes #142 

This PR removes the `return_gufe` kwarg from the `AlchemiscaleClient.query_networks` method.
As it was not a tested feature before, not tests needed to be changed.

Edit: it also doesn't appear to be mentioned in the current docs.

### DIFF
--- a/alchemiscale/interface/api.py
+++ b/alchemiscale/interface/api.py
@@ -135,25 +135,25 @@ def create_network(
 def query_networks(
     *,
     name: str = None,
-    return_gufe: bool = False,
     scope: Scope = Depends(scope_params),
     n4js: Neo4jStore = Depends(get_n4js_depends),
     token: TokenData = Depends(get_token_data_depends),
 ):
     # Intersect query scopes with accessible scopes in the token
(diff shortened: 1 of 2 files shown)


dotsdl (approved):
Looks great @ianmkenney! Merge when satisfied!

PR #207, opened 2023-12-01
Add the `visualize` kwarg to the `create_network` method

### LINKED ISSUE #156: Add `visualize` kwarg to `AlchemiscaleClient.create_network`
For consistency with other methods that produce `rich`-based output, add a `visualize` kwarg to `AlchemiscaleClient.create_network` that allows disabling of the progress output.

ianmkenney:
Fixes #156 

Added the visualize kwarg to the `create_network` method.

### DIFF
--- a/alchemiscale/interface/client.py
+++ b/alchemiscale/interface/client.py
@@ -138,7 +138,7 @@ class AlchemiscaleClient(AlchemiscaleBaseClient):
                 progress.start_task(task)
                 progress.update(task, total=1, completed=1)
         else:
-            scoped_key = get_scoped_key()
+            scoped_key = post()
 
         return ScopedKey.from_dict(scoped_key)
 
--- a/alchemiscale/interface/client.py
+++ b/alchemiscale/interface/client.py
@@ -84,6 +84,7 @@ class AlchemiscaleClient(AlchemiscaleBaseClient):
         network: AlchemicalNetwork,
         scope: Scope,
         compress: Union[bool, int] = True,
+        visualize: bool = True,
     ) -> ScopedKey:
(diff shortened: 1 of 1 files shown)


alchemiscale/interface/client.py
```diff
@@ -119,16 +122,25 @@ def create_network(
...
+            from rich.progress import Progress
+
+            with Progress(*self._rich_waiting_columns(), transient=False) as progress:
+                task = progress.add_task(
+                    f"Submitting [bold]'{sk}'[/bold]...", total=None
+                )
+
+                scoped_key = post()
+                progress.start_task(task)
+                progress.update(task, total=1, completed=1)
```
dotsdl, line 141: For reference, the reason we do `start_task` *after* we actually perform the operation is that we really just care about showing how much time has elapsed to the user as part of our indication that the client is busy doing something. We then immediately set progress to complete (`progress.update`) after "starting" it (`progress.start`) as outlined here: https://rich.readthedocs.io/en/stable/progress.html#indeterminate-progress

alchemiscale/interface/client.py
```diff
@@ -105,6 +106,8 @@ def create_network(
             Use an integer between 0 and 9 for finer control over
             the degree of compression; 0 means no compression, 9 means max
             compression. ``True`` is synonymous with level 5 compression.
+        visualize
+            If ``True``, show retrieval progress indicator.
```
dotsdl, line 110:
```suggestion
            If ``True``, show submission progress indicator.
```

dotsdl (approved):
Looks great @ianmkenney! Only one suggestion; please merge after considering it.

PR #212, opened 2023-12-05
Add client/API getter and setter methods for network weights

### LINKED ISSUE #201: Expose setting `TaskHub` weight to `AlchemiscaleClient`
Currently, it is not possible for users to set the relative `weight` of a `TaskHub` for `AlchemicalNetwork`s of interest. Increasing this `weight` increases the likelihood that the given `AlchemicalNetwork`'s actioned `Task`s are picked up by compute services relative to other `AlchemicalNetwork`s, and so this functions as a high-level mechanism for users to indicate preference for compute priority at level of entire `AlchemicalNetwork`s. Setting the `weight` to 0 also functions as a way of disabling compute on an `AlchemicalNetwork` without cancelling all of its actioned `Tasks`.

A PR closing this issue should implement the methods `get_network_weight` and `set_network_weight` on the `AlchemiscaleClient`, where weight is a float between 0.0 and 1.0. API endpoints for the `AlchemiscaleAPI` serving these client methods must be implemented to enforce this.

ianmkenney:
fixes #201 

This PR is meant to allow users to modify the task weights of tasks given a network.

### DIFF
--- a/alchemiscale/interface/client.py
+++ b/alchemiscale/interface/client.py
@@ -219,8 +219,7 @@ class AlchemiscaleClient(AlchemiscaleBaseClient):
 
     def set_network_weight(self, network: ScopedKey, weight: float) -> None:
         """Set the weight of the TaskHub associated with a given AlchemicalNetwork."""
-        data = dict(weight=weight)
-        self._post_resource(f"/networks/{network}/weight", data)
+        self._post_resource(f"/networks/{network}/weight", weight)
 
     def get_transformation_networks(self, transformation: ScopedKey) -> List[ScopedKey]:
         """List ScopedKeys for AlchemicalNetworks associated with the given Transformation."""
--- a/alchemiscale/interface/client.py
+++ b/alchemiscale/interface/client.py
@@ -213,6 +213,17 @@ class AlchemiscaleClient(AlchemiscaleBaseClient):
         """List ScopedKeys for Transformations associated with the given AlchemicalNetwork."""
         return self._query_resource(f"/networks/{network}/transformations")
 
+    def get_network_weights(self, network: ScopedKey) -> List[float]:
+        """Get the weight of actioned :class:`Task`s in an :class:`AlchemicalNetwork`"""
+        # TODO: _get_resource will return a list of weights, I assume this is more useful
+        return self._get_resource(f"/networks/{network}/weights")
+
+    def set_network_weights(self, network: ScopedKey, weight: float) -> List[ScopedKey]:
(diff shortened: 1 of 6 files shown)


alchemiscale/interface/client.py
```diff
@@ -213,6 +213,14 @@ def get_network_transformations(self, network: ScopedKey) -> List[ScopedKey]:
         """List ScopedKeys for Transformations associated with the given AlchemicalNetwork."""
         return self._query_resource(f"/networks/{network}/transformations")
 
+    def get_network_weight(self, network: ScopedKey) -> float:
+        """Get the weight of the TaskHub associated with a given AlchemicalNetwork."""
+        return self._get_resource(f"/networks/{network}/weight")
+
+    def set_network_weight(self, network: ScopedKey, weight: float) -> None:
+        """Set the weight of the TaskHub associated with a given AlchemicalNetwork."""
```
dotsdl, line 221: Let's expand these docstrings to explain what changing the `weight` does, especially what setting it to 0 does. We also need to note that `weight should be a float between 0.0 and 1.0.

Happy to discuss this point, or to contribute these docstrings to this PR myself if you prefer!

alchemiscale/tests/integration/interface/client/test_client.py
```diff
@@ -257,6 +257,46 @@ def test_get_network(
...
+        network_tyk2,
+        user_client: client.AlchemiscaleClient,
+    ):
+        an_sk = user_client.get_scoped_key(network_tyk2, scope_test)
+        with pytest.raises(
+            AlchemiscaleClientError,
+            match="Status Code 400 : Bad Request : weight must be",
+        ):
+            user_client.set_network_weight(an_sk, 1.5)
+
```
dotsdl, line 299: Also add a block where we try to set the weight to a negative float.

dotsdl (approved):
Looking great @ianmkenney! Just a few small things to address, then this is good to merge!

alchemiscale/tests/integration/interface/client/test_client.py
ianmkenney: done

alchemiscale/interface/client.py
ianmkenney: I've updated the docstring, but please feel free to change it!

dotsdl: Done! Thanks for adding these!

dotsdl (approved):
Looks great @ianmkenney! Please merge when satisfied!

PR #214, opened 2023-12-13
Add `get_network_actioned_tasks` and `get_task_actioned_networks` to `AlchemicaleClient`

### LINKED ISSUE #173: Add `get_network_actioned_tasks` and `get_task_actioned_networks` methods to `AlchemiscaleClient`
Currently, there is no exposed method in the `AlchemiscaleClient` for users to assess what `Task`s, if any, are currently actioned on a given `AlchemicalNetwork`. Users can action and cancel (de-action) `Task`s on an `AlchemicalNetwork`, but they cannot get current state.

We should add the following methods:
- `get_network_actioned_tasks`: given an `AlchemicalNetwork` `ScopedKey`, return the `ScopedKey`s for all `Task`s currently actioned on it.
- `get_task_actioned_networks`: given a `Task` `ScopedKey`, return the `ScopedKey`s for all `AlchemicalNetwork`s the `Task` is currently actioned on.

ianmkenney:
fixes #173

This PR implements methods at the `statestore`, `interface/api`, and `interface/client` levels to give users access to Tasks and `AlchemicalNetworks` based on their relationship with one another.

### DIFF
--- a/alchemiscale/interface/client.py
+++ b/alchemiscale/interface/client.py
@@ -621,17 +621,17 @@ class AlchemiscaleClient(AlchemiscaleBaseClient):
         Parameters
         ----------
         network
-            ScopedKey for an AlchemicalNetwork.
+            The ScopedKey for the AlchemicalNetwork to get actioned Tasks for.
         task_weights
-            If ``True``, return a Dict whose keys and values are the Task
-            ScopedKeys and the Task weight, respectively.
+            If ``True``, return a dict with Task ScopedKeys as keys, Task
+            weights as values.
 
         Returns
         -------
         tasks
-            A List of Task ScopedKeys. If ``task_weights`` is ``True``, then
-            the returned type is a Dict whose keys and values are the
-            Task ScopedKeys and the Task weight, respectively.
+            A list of Task ScopedKeys. If ``task_weights`` is ``True``, then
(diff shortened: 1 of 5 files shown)


alchemiscale/interface/client.py
```diff
@@ -611,6 +611,52 @@ def get_transformation_status(
...
+            return [ScopedKey.from_str(t) for t in results]
+
+    def get_task_actioned_networks(self, task: ScopedKey) -> List[ScopedKey]:
+        """Return all AlchemicalNetworks associated with a Task.
+
+        Parameters
+        ----------
+        task
+            ScopedKey of a task which may or may
+
```
dotsdl: For symmetry with `get_network_actioned_tasks`, does it make sense to also add a `task_weights` kwarg?

dotsdl (commented):
Looking good so far @ianmkenney! Thanks for working on this!

alchemiscale/interface/client.py
ianmkenney: added

dotsdl (approved):
Great work @ianmkenney! I made some major simplifications to the `AlchemiscaleAPI` methods by adding `weight`s output to the `Neo4jStore` methods they call. Let me know if you are in favor of these changes; if so, we'll merge.

Issue #107, opened 2023-03-28
Use `pydantic` models for encoding `SynchronousComputeService` settings


dotsdl:
Currently, `SynchronousComputeService` takes its intensive settings in via kwargs on its `__int__` method, with extensive settings on its `start` method (e.g. max number of tasks, max run time). This is okay, but we may be better served by creating a general settings `pydantic` model with nested `init` and `start` settings models for these two methods.

This would allow us to:
- validate inputs in the models themselves
- expand the models via inheritance for e.g. the `AsynchronousComputeService`
- keep all settings in a single object within an initialized compute service for easy export and remixing, such as in the test suite

dotsdl:
Could also use the `pydantic` model as the basis for producing a sample config file on demand via the CLI.

dotsdl:
Addressing this in #218.

PR #218, opened 2024-01-04
Changes needed to support execution via `alchemiscale-fah`

dotsdl:
Includes all functionality needed for initial support of `alchemiscale-fah`, developed here: https://github.com/openforcefield/alchemiscale-fah/pull/7

Also closes:
- #107


dotsdl:
@ianmkenney can I get a review on this one? I'll keep it [WIP] for now as we finish things out over on https://github.com/openforcefield/alchemiscale-fah/pull/7, but getting a review sooner than later here will help us address clear gaps.

alchemiscale/compute/api.py
```diff
@@ -187,13 +190,91 @@ def claim_taskhub_tasks(
...
+
+        # gather up claimed tasks, if present
+        for t in claimed_tasks:
+            if t is not None:
+                tasks.append(t)
+
+        # remove this taskhub from the options available; repeat
+        taskhubs.pop(taskhub)
+
+    return [str(t) if t is not None else None for t in tasks]
```
ianmkenney, line 253: L247 will make `t is None` impossible. The return type here could be confusing in the future.

```suggestion
    return [str(t) for t in tasks]
```

or just append `str(t)` to `tasks` on L248 and return `tasks` to avoid a needless loop.

alchemiscale/compute/api.py
```diff
@@ -187,13 +190,91 @@ def claim_taskhub_tasks(
...
+    *,
+    n4js: Neo4jStore = Depends(get_n4js_depends),
+    token: TokenData = Depends(get_token_data_depends),
+):
+    sk = ScopedKey.from_str(task_scoped_key)
+    validate_scopes(sk.scope, token)
+
+    transformation: ScopedKey
+
+    transformation, protocoldagresultref = n4js.get_task_transformation(
```
ianmkenney, line 268: Unused variable

```suggestion
    transformation, _ = n4js.get_task_transformation(
```

alchemiscale/compute/api.py
```diff
@@ -187,13 +190,91 @@ def claim_taskhub_tasks(
         taskhub=taskhub_scoped_key,
         compute_service_id=ComputeServiceID(compute_service_id),
         count=count,
+        protocols=protocols,
```
ianmkenney, line 193: Note that `n4js.claim_taskhub_tasks` has protocols typed as `Protocol`s, not `str`.

pyproject.toml
```diff
@@ -76,3 +76,7 @@ distance-dirty = "{base_version}+{distance}.{vcs}{rev}.dirty"
 method = "git" 
 match = ["*"]
 default-tag = "0.0.0"
+
+
+#[project.entry-points.pytest11]
```
ianmkenney, line 81: Do we plan on uncommentating these?

alchemiscale/compute/client.py
```diff
@@ -71,19 +73,48 @@ def query_taskhubs(
...
+        """Claim Tasks from TaskHubs within a list of Scopes."""
+        data = dict(
+            scopes=[scope.dict() for scope in scopes],
+            compute_service_id=str(compute_service_id),
+            count=count,
+            protocols=protocols,
+        )
+        tasks = self._post_resource("/claim", data)
 
         return [ScopedKey.from_str(t) if t is not None else None for t in tasks]
```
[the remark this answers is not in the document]
dotsdl: Ah, this is an oversight on my part. Ideally we'd like the API call to always given back a list of length `count`, even if some elements must be `None`. This is to mimic the behavior of the older `claim_taskhub_tasks` method.

I've adjusted the return statement of the compute API `claim_tasks` endpoint to reflect this.

alchemiscale/compute/api.py
dotsdl: Thanks for your fixes in #279!

pyproject.toml
dotsdl: No, I'll cut them now. They were an experiment for exposing some of our fixtures via entrypoints so we could use them for e.g. `alchemiscale_fah`, but I did this in a different way instead.

dotsdl:
Thanks so much @ianmkenney! I'm going to merge; any additional changes made to support `alchemiscale-fah` will come in subsequent PRs.

Issue #216, opened 2023-12-21
`AlchemiscaleClient.create_network` scales poorly with increasing `AlchemicalNetwork` size


dotsdl:
Currently, the `AlchemiscaleClient.create_network` method calls [`network.to_dict()`](https://github.com/openforcefield/alchemiscale/blob/8e98af36fb8847f909600f92ba0d95172d4d4124/alchemiscale/interface/client.py#L128) before serializing this to JSON and shipping it to the `AlchemiscaleAPI`. This is an inefficient approach that scales very poorly with increasing `AlchemicalNetwork` size, as any `Component`s used by the network are duplicated many times in the resulting object, including potentially large `ProteinComponent`s.

Instead of doing this, we should make use of `network.to_keyed_dict`, and use a method similar to [`gufe.tokenization.get_all_gufe_objs`](https://github.com/OpenFreeEnergy/gufe/blob/ae96d30952a4e000fcb690443a6dbd79ec31756a/gufe/tokenization.py#L804) for decomposing the `AlchemicalNetwork` into its component keyed-dicts, ordered in such a way that they could all be re-assembled, and serializing+shipping these to the `AlchemiscaleAPI` instead.

Ideally, this same approach could be used for other `AlchemiscaleClient` methods to improve their efficiency, such as `AlchemicalNetwork` retrieval.

PR #219, opened 2024-01-06
Optimization to `gufe_to_digraph` to avoid repeated traversals, calls to `GufeTokenizable.to_shallow_dict`

dotsdl:
Builds on the solution to #216.

Added use of a `shallow_dicts` index to avoid potentially expensive repeated calls of `GufeTokenizable.to_shallow_dict`, such as for `ProteinComponent`s. Also used as our check if we've already processed the object being visited to avoid going down its dependency tree again, avoiding repeated traversals.

For an `AlchemicalNetwork` with ~1000 `ChemicalSystem`s and ~2000 `Transformation`s, this optimization reduces `gufe_to_digraph` execution time from **6min 31s** to **811ms**.

### DIFF
--- a/alchemiscale/utils.py
+++ b/alchemiscale/utils.py
@@ -84,20 +84,36 @@ def gufe_to_digraph(gufe_obj):
     other GufeTokenizables.
     """
     graph = nx.DiGraph()
+    shallow_dicts = {}
 
-    def add_edges(o):
+    def add_edges(o, sd):
         # add the object node in case there aren't any connections
         graph.add_node(o)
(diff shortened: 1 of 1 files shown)


ianmkenney (approved):
I think we could move a lot of the logic added to `modifier` to `add_edges` instead, allowing `add_edges` to check if the shallow dict has already been processed. On the other hand, keeping `add_edges` smaller makes sense (at the expense of code duplication).

By combining you could remove one of the 

```python
sd = gufe_obj.to_shallow_dict()
shallow_dicts[gufe_obj.key] = sd
```

blocks.

Either way works for me, though.

PR #224, opened 2024-01-09
Attempting to fix seemingly random failures on CI

dotsdl:
Unfortunately, unable to reproduce failures like [this](https://github.com/openforcefield/alchemiscale/actions/runs/7467107404/job/20319915813#step:6:420) locally. Working hypothesis is that use of fixtures is impacting us in hard to pin-down ways, so scoping them down gradually here.

### DIFF
--- a/alchemiscale/interface/client.py
+++ b/alchemiscale/interface/client.py
@@ -76,7 +76,7 @@ class AlchemiscaleClient(AlchemiscaleBaseClient):
                 "Scope for a ScopedKey must be specific; it cannot contain wildcards."
             )
 
-    def check_exists(self, scoped_key: Scope) -> bool:
+    def check_exists(self, scoped_key: ScopedKey) -> bool:
         """Returns ``True`` if the given ScopedKey represents an object in the database."""
         return self._get_resource(f"/exists/{scoped_key}")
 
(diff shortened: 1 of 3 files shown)


dotsdl:
After much gnashing of teeth, I think I may have narrowed down the cause of the random CI failures to a race condition. Hypothesis that has yet to be invalidated:
- following the call to [`user_client.create_network`](https://github.com/openforcefield/alchemiscale/blob/518ea886d940c8c60492851439c0ce53413a43db/alchemiscale/tests/integration/interface/client/test_client.py#L1336) completes, but then when [`user_client.get_network_transformations`](https://github.com/openforcefield/alchemiscale/blob/518ea886d940c8c60492851439c0ce53413a43db/alchemiscale/tests/integration/interface/client/test_client.py#L1341) is called later, no `Transformation` `ScopedKey`s are returned, suggesting that the query to server doesn't see them yet
- adding a `while` loop to call `get_network_transformations` repeatedly until `Transformation`s are populated appears to address the issue
- it remains unclear how this race condition could be present; it may be the case that the calls to Neo4j to create the `AlchemicalNetwork` via `py2neo` fully return before Neo4j has finished creating all nodes and relationships in the DB, and the low amount of total processing allocated on the CI worker presents conditions such that this creation may be incomplete before the next call in the tests occurs

The solution to this may be to try pulling the full `AlchemicalNetwork`, and do this until it succeeds, then progress to the remaining tests. A note on why this is necessary would be sufficient.

dotsdl:
6 successful attempts in a row! I think we might be good. :grin:

PR #227, opened 2024-01-13
Adding KeyedChain class

ianmkenney:
The KeyedChain class enhances the usage of the keyed_dict representation recently implemented in #217 and #219. The new class contains common methods to be used with the structure.

### DIFF
--- a/alchemiscale/tests/unit/test_keyedchain.py
+++ b/alchemiscale/tests/unit/test_keyedchain.py
@@ -1,0 +1,46 @@
+from alchemiscale.keyedchain import KeyedChain
+from alchemiscale.utils import RegistryBackup
+
+from gufe.tokenization import get_all_gufe_objs, GufeTokenizable, is_gufe_key_dict
+
+
+def test_keyedchain_full_network(network):
+    objects = get_all_gufe_objs(network)
+
+    for o in objects:
+        with RegistryBackup():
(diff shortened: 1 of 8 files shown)


alchemiscale/tests/unit/test_keyedchain.py
```diff
@@ -0,0 +1,46 @@
...
+
+    assert len(keyedchain) == expect_len
+
+
+def test_keyedchain_get_keys(network):
+    keyedchain = KeyedChain.from_gufe(network)
+    keys = list(map(lambda x: x.key, get_all_gufe_objs(network)))
+
+    for key in keyedchain.gufe_keys():
+        assert key in keys
```
dotsdl, line 31: Consider instead doing a `set` comparison, since as written this only checks that all keys in the `KeyedChain` are in `keys`, but not the inverse.

dotsdl (approved):
Looking great @ianmkenney! Please address the comments, then merge when satisfied!

ianmkenney:
The `TOKENIZABLE_REGISTRY` is still updated with the intermediate objects, but isn't used for constructing the target `GufeTokenizable`. Therefore, if the hashing process changes between gufe versions, the original KeyedChain and a newly generated KeyedChain will not be the same.

dotsdl (approved):
Awesome work @ianmkenney! Merging!

PR #235, opened 2024-02-13
remove py2neo as primary database driver

ianmkenney:
We currently use the EOL `py2neo` library for handling communication between `alchemiscale` and the Neo4j database. This PR removes this responsibility from `py2neo` and instead shifts it to the official Neo4j Python driver. This PR does *not* remove py2neo entirely since we rely heavily on the `py2neo` `Subgraph` data structure and query generation.


dotsdl (approved):
Amazing work @ianmkenney! Merge if you're satisfied!

Issue #236, opened 2024-02-13
allow setting of `task_status` to "error"


JenkeScheen:
We're building a feature in `asap-alchemy` that will allow us to `stop` networks that have tasks `running` or `waiting`. This is useful in cases where we don't care about finishing a network all the way until the last few tasks have finished because we have already collected meaningful results for a prediction deadline. 

We can do this using `alchemiscale.interface.client.set_tasks_status()` after collecting the network's `running`/`waiting` tasks. However, calling `alchemiscale.interface.client.set_tasks_status(tasks, "error")` results in a non-descriptive server-side 500:
```
$ asap-alchemy stop -nk AlchemicalNetwork-e7a[redacted]
Traceback (most recent call last):
  File "/opt/homebrew/Caskroom/miniconda/base/envs/asapdiscovery/bin/asap-alchemy", line 8, in <module>
    sys.exit(alchemy())
  File "/opt/homebrew/Caskroom/miniconda/base/envs/asapdiscovery/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/asapdiscovery/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/asapdiscovery/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/homebrew/Caskroom/miniconda/base/envs/asapdiscovery/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/homebrew/Caskroom/miniconda/base/envs/asapdiscovery/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/Users/jenkescheen/asapdiscovery/asapdiscovery-alchemy/asapdiscovery/alchemy/cli/alchemy.py", line 487, in stop
    deleted = [ sc_k for sc_k in client._client.set_tasks_status(running + waiting, "error"
  File "/opt/homebrew/Caskroom/miniconda/base/envs/asapdiscovery/lib/python3.10/site-packages/alchemiscale/interface/client.py", line 891, in set_tasks_status
    return self._task_attribute_setter(
[...]

PR #238, opened 2024-02-14
Fixes a bug identified by @JenkeScheen in #236

dotsdl:
`status` was a module import, but was being overriden by the use of `status` as an arg in certain API endpoints.

This PR fixes this issue.


dotsdl:
Can I get a review on this one @ianmkenney?

alchemiscale/compute/api.py
```diff
@@ -9,7 +9,8 @@
 import json
 from datetime import datetime, timedelta
 
-from fastapi import FastAPI, APIRouter, Body, Depends, HTTPException, status
+from fastapi import FastAPI, APIRouter, Body, Depends, HTTPException
+from fastapi import status as http_status
```
ianmkenney, line 13: Currently, `compute.api` doesn't use `HTTPException` or `status`. We can just remove them.

ianmkenney, line 13:
```suggestion
from fastapi import FastAPI, APIRouter, Body, Depends
```

ianmkenney:
Other than the unused imports, seems sensible to me

alchemiscale/compute/api.py
dotsdl: We should start using them in various places to give more descriptive error messages, so I'd rather leave them in as a kind of placeholder.

dotsdl: Accepted your suggestion; can proceed later with adding these back in as needed.

PR #242, opened 2024-02-20
Support Neo4j 5.x

ianmkenney:
This PR allows to use of Neo4j 5.x with `alchemiscale`.

PR #244, opened 2024-02-22
Optimizations for TaskHub task claiming

### LINKED ISSUE #121: Compute-side bottleneck when many compute services attempt to claim from a single `TaskHub`
First reported [here](https://github.com/openforcefield/alchemiscale/pull/120#issuecomment-1510511650) in #120.

We observe a bottleneck compute-side with task claiming when there are a small number of `TaskHub`s/`AlchemicalNetwork`s with available `Task`s, since the claiming code must currently lock the `TaskHub` while performing claiming. This linearizes requests for new `Task`s from a given `TaskHub` by compute services, causing a bit of starving compute.

This can be mitigated by setting the number of tasks claimed at a time on compute services higher. The problem will also be diminished when many `TaskHub`s have waiting tasks, as well as when tasks take much longer than the ~1min tasks I've been testing with.

Addressing this issue directly will require a more efficient claiming code path. Ideally we would be able to claim `Task`s from a `TaskHub` in a single Cypher query, but this may not be possible given our approach to claiming. Alternative to this, a rework of the claiming code that avoids the need to lock the `TaskHub` being claimed from may also yield higher aggregate performance.

ianmkenney:
This PR optimizes task claiming by putting more burden on Neo4j for priority sorting, removing the use of Subgraphs, intelligently stopping early, and generally sending fewer queries across the entire process. This closes #121 and should be reviewed after #242 is merged.


alchemiscale/storage/statestore.py
```diff
@@ -96,15 +97,29 @@ def _select_task_from_taskpool(taskpool: Subgraph) -> Union[ScopedKey, None]:
     return chosen_one[0]
 
 
+def _select_tasks_from_taskpool(taskpool: List[Tuple[str, float]], count) -> List[str]:
```
dotsdl, line 100: Since we have this, can we delete `_select_task_from_taskpool` above?

dotsdl, line 100: Add a small docstring.

dotsdl (approved):
Looks great @ianmkenney! Fantastic work on this!

Just one note on removing dead code, and I added clarifying comments where I thought they would help us in the future if we have to revisit this.

dotsdl:
Once you've made changes you think are appropriate, please merge!

alchemiscale/storage/statestore.py
ianmkenney: Removed

ianmkenney: Added docstring

alchemiscale/storage/statestore.py
```diff
@@ -136,14 +136,15 @@ def _generate_claim_query(
...
     WHERE NOT (t)<-[:CLAIMS]-(:ComputeServiceRegistration)
-    SET t.status = 'running'
 
     WITH t
 
     // create CLAIMS relationship with given compute service
     MATCH (csreg:ComputeServiceRegistration {{identifier: '{compute_service_id}'}})
     CREATE (t)<-[cl:CLAIMS {{claimed: localdatetime('{datetime.utcnow().isoformat()}')}}]-(csreg)
 
+    SET t.status = 'running'
```
ianmkenney: updating this to use `TastStatusEnum.running.value`

ianmkenney:
Merging on CI pass

alchemiscale/storage/statestore.py
dotsdl: Awesome, thank you!

PR #255, opened 2024-03-11
Add more bulk methods to the Alchemiscale client

### LINKED ISSUE #243: Add `get_networks_weight` and `set_networks_weight` methods as bulk methods to `AlchemiscaleClient`
From #239, there is a desire from users to be able to set the relative `weight`s of many `AlchemicalNetwork` at once, instead of one at a time via multiple calls to `get_network_weight` and `set_network_weight`. This could be accomplished with a pattern similar to that used for `get_tasks_status` and `set_tasks_status`.

We propose adding methods `get_networks_weight` and `set_networks_weight` to the `AlchemiscaleClient` to support bulk operations, using a similar `/bulk/networks/weight` API endpoint and use of batching + `asyncio` to make this relatively efficient for large numbers of `AlchemicalNetwork`s.

### LINKED ISSUE #251: Add `get_networks_status` and `get_networks_actioned_tasks` as bulk methods to `AlchemiscaleClient`
Similarly to #243, users are interested in being able to get `Task` status information and actioned `Task`s for multiple `AlchemicalNetwork`s at once, instead of one at a time using the existing `get_network_status` and `get_network_actioned_task`s methods.

We propose adding the following bulk methods:
- `get_networks_status`: takes a list of `AlchemicalNetwork` `ScopedKey`s, returns a list of the same length giving `dict` of status counts of the same form as `get_network_status`
- `get_networks_actioned_tasks`: takes a list of `AlchemicalNetwork` `ScopedKey`s, returns a list of the same length giving `list`s of `Task` `ScopedKey`s of the same form as `get_network_actioned_tasks`

This would specifically improve performance in `asapdiscovery-alchemy`: https://github.com/choderalab/asapdiscovery/issues/899

ianmkenney:
This PR introduces 

- [x] `get_networks_status`
- [x] `get_networks_actioned_tasks`
- [x] `get_networks_weight`
- [x] `set_networks_weight`

methods to the Alchemiscale client. The first two close #251 and the other two close #243.

### DIFF
--- a/alchemiscale/interface/client.py
+++ b/alchemiscale/interface/client.py
@@ -623,6 +623,14 @@ class AlchemiscaleClient(AlchemiscaleBaseClient):
 
         return status_counts
 
+    def get_networks_status(
+        self,
+        networks: List[ScopedKey],
+    ) -> List[Dict[str, int]]:
+        data = {"networks": [str(network) for network in networks]}
+        status_counts = self._post_resource("/bulk/networks/status", data=data)
+        return status_counts
+
     def get_transformation_status(
         self,
         transformation: ScopedKey,
@@ -679,6 +687,25 @@ class AlchemiscaleClient(AlchemiscaleBaseClient):
             return {ScopedKey.from_str(t): w for t, w in tasks.items()}
 
         return [ScopedKey.from_str(t) for t in tasks]
+
+    def get_networks_actioned_tasks(
+        self,
+        networks: List[ScopedKey],
+        task_weights: bool = False,
+    ) -> List[Union[Dict[ScopedKey, float], List[ScopedKey]]]:
+        data = dict(
+            networks=[str(network) for network in networks], task_weights=task_weights
+        )
+        grouped_tasks = self._post_resource("/bulk/networks/tasks/actioned", data=data)
+
+        return_data = []
+        for tasks in grouped_tasks:
+            if task_weights:
+                return_data.append({ScopedKey.from_str(t): w for t, w in tasks.items()})
+            else:
+                return_data.append([ScopedKey.from_str(t) for t in tasks])
+
+        return return_data
 
     def get_task_actioned_networks(
         self, task: ScopedKey, task_weights: bool = False
--- a/alchemiscale/interface/client.py
+++ b/alchemiscale/interface/client.py
@@ -228,7 +228,7 @@ class AlchemiscaleClient(AlchemiscaleBaseClient):
 
     def get_networks_weight(
         self, networks: List[ScopedKey], batch_size: int = 1000
-    ) -> [float]:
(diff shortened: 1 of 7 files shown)


dotsdl:
Reviewing this today!

alchemiscale/interface/client.py
```diff
@@ -243,6 +257,25 @@ def set_network_weight(self, network: ScopedKey, weight: float) -> None:
...
+    async def _set_network_weight(self, networks: List[ScopedKey], weight) -> None:
+        data = dict(networks=[str(network) for network in networks], weight=weight)
+        await self._post_resource_async("/bulk/networks/weight/set", data=data)
+
+    def set_networks_weight(
+        self,
+        networks: List[ScopedKey],
+        weight: float,
+        batch_size: int = 1000,
+    ) -> None:
```
dotsdl, line 269: Actually, want this to return the `ScopedKey`s of `AlchemicalNetwork`s that were updated, in the same order as given in `networks`. If a given `AlchemicalNetwork` doesn't exist, ``None`` returned in its place.

This is consistent with how e.g. `set_tasks_status` works.

alchemiscale/interface/client.py
```diff
@@ -243,6 +257,25 @@ def set_network_weight(self, network: ScopedKey, weight: float) -> None:
...
         self._post_resource(f"/networks/{network}/weight", weight)
 
+    async def _set_network_weight(self, networks: List[ScopedKey], weight) -> None:
+        data = dict(networks=[str(network) for network in networks], weight=weight)
+        await self._post_resource_async("/bulk/networks/weight/set", data=data)
+
+    def set_networks_weight(
+        self,
+        networks: List[ScopedKey],
+        weight: float,
```
dotsdl, line 267: This might be challenging to do, but I think supporting `weight` as a list of floats equal in length to `networks` is desirable, as this would allow users to use `get_networks_weight` to get the weights for many networks, alter them, then use `set_networks_weight` with these altered weights to change them in one call.

What do you think @ianmkenney?

dotsdl (changes requested):
Looking great @ianmkenney! Thanks for all this additions!

I do have some notes on changes to make; please address these when you get the chance!

alchemiscale/interface/client.py
ianmkenney: The original `get_network_weight` returns a `None`. Do you want me to change that, too?

dotsdl: I suppose so, go for it!

dotsdl (approved):
Looks great @ianmkenney! I made some changes to enable setting individual weights per `AlchemicalNetwork` with a single `AlchemiscaleClient.set_networks_weight` call; please feel free to check my work here and merge when satisfied!

PR #256, opened 2024-03-12
Added DeprecationWarning to AlchemiscaleClient.get_scoped_key

### LINKED ISSUE #248: Deprecate `AlchemiscaleClient.get_scoped_key`
Early on, we created the `AlchemiscaleClient.get_scoped_key` method as a convenience for users to construct `ScopedKey`s from a `GufeTokenizable` and a `Scope`. However, in practice using this method doesn't give a reliable way to construct `ScopedKey`s for objects represented server-side, as the `gufe` key of e.g. a `Transformation` locally may not be exactly the same as for the same `Transformation` server-side.

Since we now have methods for getting the actual `ScopedKey`s of objects server-side, such as `AlchemiscaleClient.get_network_transformations`, we propose deprecating `AlchemiscaleClient.get_scoped_key`.

Since we use it heavily in the test suite, we can move it to elsewhere in the library as a standalone function.

ianmkenney:
Closes #248

### DIFF
--- a/alchemiscale/interface/client.py
+++ b/alchemiscale/interface/client.py
@@ -32,6 +32,8 @@ from ..security.models import CredentialedUserIdentity
 from ..validators import validate_network_nonself
 from ..keyedchain import KeyedChain
 
+from warnings import warn
+
 
 class AlchemiscaleClientError(AlchemiscaleBaseClientError): ...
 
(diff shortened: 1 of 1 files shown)


dotsdl (approved):
Looks great @ianmkenney! Feel free to merge when satisfied!

Issue #123, opened 2023-04-17
Occasionally see `running` `Task`s without a `CLAIMS` relationship to a `ComputeServiceRegistration`


dotsdl:
After some execution time, with compute service instances killed at random times, we've seen some `Task`s with status `running` despite having now `CLAIMS` relationship to a `ComputeServiceRegistration`. 

It is unclear under what conditions this can happen, as the deregistration and expiration queries should reset the status to `waiting` before the claiming `ComputeServiceRegistration` is removed along with its relationships.

dotsdl:
This isn't critical in most cases, as users can always create new tasks or reset these `running` tasks to `waiting` themselves, but it is worth identifying and resolving the root cause.

dotsdl:
Perhaps we need to be doing a [FOREACH](https://neo4j.com/docs/cypher-manual/4.4/clauses/foreach/) when setting task status back to `waiting` in the deregister/expire methods?

dotsdl:
Seeing this repeatedly occur on many tasks on the alchemiscale.org instance. Will use this opportunity to try to identify the underlying issue so we can resolve this one.

dotsdl:
Observed the following in `alchemiscale.org` logs when this problem occurs:
```
docker-alchemiscale-compute-API-1  | RuntimeError: coroutine raised StopIteration
docker-alchemiscale-compute-API-1  | [2024-03-26 04:51:34 +0000] [42] [ERROR] Exception in ASGI application
docker-alchemiscale-compute-API-1  | Traceback (most recent call last):
docker-alchemiscale-compute-API-1  |   File "/opt/conda/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 807, in run
docker-alchemiscale-compute-API-1  |     result = context.run(func, *args)
docker-alchemiscale-compute-API-1  |   File "/tmp/alchemiscale/compute/api.py", line 122, in deregister_computeservice
docker-alchemiscale-compute-API-1  |     compute_service_id_ = n4js.deregister_computeservice(
docker-alchemiscale-compute-API-1  |   File "/tmp/alchemiscale/storage/statestore.py", line 888, in deregister_computeservice
docker-alchemiscale-compute-API-1  |     identifier = next(res)["identifier"]
docker-alchemiscale-compute-API-1  |   File "/opt/conda/lib/python3.10/site-packages/py2neo/cypher/__init__.py", line 110, in __next__
[...]

PR #260, opened 2024-03-27
Added a test to check that ComputeService deregistration resets claimed Tasks to status waiting

dotsdl:
This is intended as a sanity check on #123, since we haven't pinned down the conditions under which this occurs, but weren't explicitly testing that deregistration resets Task status to waiting.


dotsdl:
@ianmkenney really small PR; can I get a review from you at your convenience?

ianmkenney (approved):
Seems sensible, should be good to merge!

PR #269, opened 2024-04-16
More HTTPExceptions in the interface API endpoints

### LINKED ISSUE #163: Add `HTTPException`s for object and result retrieval in `AlchemiscaleAPI`
Currently, if a user requests an object that does not exist via the `AlchemiscaleClient`, they are likely to get an HTTP error with status code "500: Internal Server Error". 

This is opaque to users, and so instead we should `try...except` retrievals in all `AlchemiscaleAPI` endpoints and raise `HTTPException`s with more descriptive detail on what was attempted and what occurred.

ianmkenney:
Closes #163. This PR introduces more HTTPExceptions in the interface API endpoints. Tests hitting these Exceptions were added when the implementations were obvious.

### DIFF
--- a/alchemiscale/interface/api.py
+++ b/alchemiscale/interface/api.py
@@ -312,6 +312,7 @@ def get_chemicalsystem_transformations(
     ]
 
 
+# TODO: HTTPException
 @router.get("/networks/{network_scoped_key}")
 def get_network(
     network_scoped_key,
@@ -326,6 +327,7 @@ def get_network(
     return GufeJSONResponse(network)
 
 
+# TODO: HTTPException
 @router.get("/transformations/{transformation_scoped_key}")
 def get_transformation(
     transformation_scoped_key,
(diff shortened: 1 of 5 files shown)


dotsdl (approved):
Looks great @ianmkenney! Fantastic work!

I added an additional check to `Neo4jStore.get_taskhub`, and adjusted the API endpoints accordingly to account for this.

Feel free to merge when satisfied!

alchemiscale/interface/api.py
```diff
@@ -614,9 +614,8 @@ def get_network_actioned_tasks(
     validate_scopes(network_sk.scope, token)
 
     try:
-        # raises ValueError when a ScopedKey doesn't correspond to an AlchemicalNetwork
         taskhub_sk = n4js.get_taskhub(network_sk)
-    except ValueError as e:
+    except Exception as e:
```
ianmkenney, line 618: @dotsdl should we instead `except (ValueError, KeyError)` given your change?

dotsdl: Ah good idea! This is better yes!

PR #270, opened 2024-04-17
Added support for `NonTransformation`s

### LINKED ISSUE #110: Add explicit `NonTransformation` support throughout codebase
We currently operate on `Transformation`s throughout `alchemiscale`, but `AlchemicalNetwork`s may also feature `NonTransformation`s, which are self-loops on a node (the same node is on both ends of the edge). The way the state store and other components are written, currently `NonTransformation`s are not fully supported. It may be possible to submit `AlchemicalNetwork`s featuring `NonTransformation`s, but creating `Task`s on them will likely not work.

We should add in explicit support for `NonTransformation`s in every component applicable, especially the state store.

ianmkenney:
This PR closes #110. The default `network_tyk2` fixture now includes `NonTransformation`s on all `ChemicalSystem`s. All Cypher queries involving `Transformation`s now also include `NonTransformation`s. Self loop validation now only runs when the transformation is of type `Transformation`.

### DIFF
--- a/alchemiscale/storage/statestore.py
+++ b/alchemiscale/storage/statestore.py
@@ -689,7 +689,7 @@ class Neo4jStore(AlchemiscaleStateStore):
         # not yet tested though
         """
         MATCH p = (n:AlchemicalNetwork {{_scoped_key: "{network_node['_scoped_key']}")-[r:DEPENDS_ON*]->(m),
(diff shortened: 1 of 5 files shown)


dotsdl (approved):
Looks great @ianmkenney! Just one note; merge when satisfied!

PR #274, opened 2024-05-07
Add espaloma_charge and nagl

mikemhenry:
Might need to play with things a bit to make sure this works, will ping when ready for review/merge


mikemhenry:
Image here: `mmh42/alchemiscale-mike:test` for testing, I don't have time to spin up a worker with it right now

PR #275, opened 2024-05-13
Fix regular expression matching in `query_networks`

### LINKED ISSUE #273: `Neo4jStore.query_networks` doesn't yield `AlchemicalNetwork`s with `name = None`
With the addition of `NetworkMark` in #257, we changed `Neo4jStore.query_networks` to perform regex-based matches on attributes. A consequence of this is that if an `AlchemicalNetwork` has no `name` attribute (`name = None`), then `query_network`s will not return its `ScopedKey` at all, rendering it invisible to users. The `AlchemicalNetwork` is still referencable directly with its `ScopedKey`, however.

To fix this, we propose changing the [query we use](https://github.com/openforcefield/alchemiscale/blob/0954e32573d157271e565f488ccf72a61821bd58/alchemiscale/storage/statestore.py#L864-L872) to a query-builder approach, where we only populate `AND` clauses in the `WHERE` clause where an attribute is not `None`. This should avoid the issue where we assume `Alchemiscale.name` is always a string.

ianmkenney:
Since attributes of `GufeTokenizable`s that are set to None are not explicitly stored in the Neo4j database, regular expression fails to match nodes. Because of this, we no longer introduce wildcards when a attribute pattern is set to None. Instead, we completely remove that attribute regex condition from the WHERE clause.

The test for the `query_networks` method was also changed so a network without a name is included.

Closes #273

### DIFF
--- a/alchemiscale/storage/statestore.py
+++ b/alchemiscale/storage/statestore.py
@@ -856,19 +856,26 @@ class Neo4jStore(AlchemiscaleStateStore):
             gufe_key_pattern=None if key is None else str(key),
         )
 
-        for k, v in query_params.items():
-            if v is None:
-                query_params[k] = ".*"
+        where_params = dict(
+            name_pattern="an.name",
+            org_pattern="an.`_org`",
+            campaign_pattern="an.`_campaign`",
+            project_pattern="an.`_project`",
+            state_pattern="nm.state",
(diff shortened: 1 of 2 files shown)


dotsdl (approved):
Great work @ianmkenney! Thank you for solving this one!!

PR #254, opened 2024-03-05
Test new openfe & gufe rc

mikemhenry:

PR #276, opened 2024-05-30
Fix for create_network in interface API, properly deserializing AlchemicalNetwork

ianmkenney:
Directly handle the request to `create_network` using the gufe `JSON_HANDLER` to properly decode the contained `Settings` objects.

Since these changes also work with `gufe < 1.x.x`, this PR shouldn't be blocked by #254.

### DIFF
--- a/alchemiscale/interface/api.py
+++ b/alchemiscale/interface/api.py
@@ -7,9 +7,12 @@
 from typing import Dict, List, Optional, Union
 from collections import Counter
 
-from fastapi import FastAPI, APIRouter, Body, Depends, HTTPException
+from fastapi import FastAPI, APIRouter, Body, Depends, HTTPException, Request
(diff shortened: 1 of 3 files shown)


dotsdl:
> @dotsdl There was a bug in the preloaded database fixture. At first it appeared random (like the supposed race conditions we have seen before). Upon further inspection I couldn't figure out how it could have been. Rerunning the tests **many** times locally, I finally caught one in the debugger. Long story short, we were trying to actions tasks for transformations that were not part of the alchemical network and our failing test relied on all tasks being actioned to pass.
> 
> image

Great detective work on this @ianmkenney! Thanks for fixing this bug in the test suite!

dotsdl (approved):
Looks excellent @ianmkenney! Thank you!

PR #278, opened 2024-07-11
Increment compute env to use cudatoolkit 11.8

dotsdl:
We've experienced frequent errors with using OpenMM 8.1.1:

    openmm.OpenMMException: Error initializing FFT: 5

and the culprint may be `cudatoolkit` 11.7: https://github.com/pytorch/pytorch/issues/88038

In an attempt to address this, pinning to the latest `cudatoolkit` available on `conda-forge` as of this writing (11.8) to see if these errors drop in frequency.


dotsdl:
This appears to be yielding positive results in our use on MSK Iris; merging!

PR #279, opened 2024-07-11
Test TaskHub Task claiming by Protocol

ianmkenney:
* Minor changes to new claiming code (cypher query generation fails when provided a `Protocol` instance)
* Introduce a new set of `DummyProtocol` subclasses
* Test `Task` claiming with a `Protocol` list

### DIFF
--- a/alchemiscale/tests/integration/storage/test_statestore.py
+++ b/alchemiscale/tests/integration/storage/test_statestore.py
@@ -1324,6 +1324,55 @@ class TestNeo4jStore(TestStateStore):
         claimed6 = n4js.claim_taskhub_tasks(taskhub_sk, csid, count=2)
         assert claimed6 == [None] * 2
 
+    def test_claim_taskhub_tasks_protocol_split(
+        self, n4js: Neo4jStore, network_tyk2, scope_test
+    ):
+        an = network_tyk2
+        network_sk, taskhub_sk, _ = n4js.assemble_network(an, scope_test)
+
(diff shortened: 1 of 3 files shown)


alchemiscale/tests/integration/storage/test_statestore.py
```diff
@@ -1324,6 +1327,52 @@ def test_claim_taskhub_tasks(self, n4js: Neo4jStore, network_tyk2, scope_test):
...
+                return collection
+            sk = n4js.get_scoped_key(transformation, scope_test)
+            collection[transformation.protocol.__class__].append(sk)
+            return collection
+
+        transformations = reduce(
+            reducer,
+            an.edges,
+            {DummyProtocolA: [], DummyProtocolB: [], DummyProtocolC: []},
+        )
```
dotsdl, line 1348: Would this all be about the same as using a `defaultdict(list)` and adding each `Transformation` `ScopedKey` up to a max count for each key? This form is fine, but I probably would have expressed it with a `defaultdict` and loop instead.

dotsdl (approved):
Looks great @ianmkenney! One clarifying comment, but this is good to merge if you're satisfied!

dotsdl:
Thank you so much for doing this!

alchemiscale/tests/integration/storage/test_statestore.py
dotsdl: Happy to keep it! You taught me something new as well; wanted to make sure I was understanding it right. :grin:

PR #283, opened 2024-07-23
release 0.5.0 prep

dotsdl:
- **Update prod envs to use Python 3.11**
- **Update all envs to use gufe 1.0.0, openfe 1.0.1**


mikemhenry:
Don't forget to add `curl` to the env

dotsdl:
Seeing no issues with Python 3.12 use in my testing. @mikemhenry, @IAlibay, @j-wags can you remind me what reservations we had for using it?

IAlibay:
I think it was ambertools / openmmforcefields but that's fixed now. At least OpenFE runs CI on 3.12 just fine. 

Only dependency I'm not sure about is oechem, since we don't use it that much in our tests.

dotsdl:
Okay cool, this is great news! I think we can proceed with Python 3.12 and see if we actually hit any issues from users.

PR #288, opened 2024-08-14
Switch pip installs to conda packages where possible, add restart policies to docker-compose containers

dotsdl:
A small set of adjustments from deploying v0.5.0.


dotsdl:
@mikemhenry could I get a review from you on this? Basically looking for any problems with what I've chosen to do here.

ianmkenney (approved):

Issue #292, opened 2024-08-20
Update `alchemiscale` test suite to use `neo4j` 5.23


dotsdl:
We currently test against `neo4j` 5.16 in the test suite; we should up this to at lest 5.22, which we are now using in production deployments.

ianmkenney:
Notes from working session

``` 
neo4j version  fail?  reason                                                  
-------------------------------------------------------------------------------
          5.16  no     --                                                      
          5.17  no     --                                                     
          5.18  no     --                                                      
          5.19  yes    ValueError, errors.py 56, split code fails (code was None) 
          5.20  yes    ValueError, errors.py 56, split code fails (code was None) 
          5.21  yes    ValueError, errors.py 56, split code fails (code was None) 
          5.22  yes    ValueError, errors.py 56, split code fails (code was None) 
          5.23  yes    ClientError, http.py 482, rs.audit, _hello                 
          5.24  yes    ClientError, http 482, rs.audit, _hello                 
          5.25  yes    ClientError, http 482, rs.audit, _hello                 
```

1 Code split error
==================

  Note that the code split error occurs within Neo4j.split_code method
  in `py2neo/errors.py'.

  We are going to just start from 5.25 so we'll revisit this if necessary. 


2 ClientError
=============

  Dropped into pdb error returns a message "not found" from neo4j
  database

  request went out to /db/data which no longer exists at the py2neo layer. Switching to
  /db/neo4j/tx with POST method fixed the problem and tests passed.

  Go into `grolt/__init__.py` and rewrite `Neo4jMachine.ping`. Remove use of
  py2neo.

  TODO: implement `_poll_http` and `_poll_bolt` using the official neo4j driver

dotsdl:
This PR will fix this: https://github.com/datryllic/grolt/pull/1

Issue #291, opened 2024-08-20
Deprecation warnings from `neo4j`: "The query used a deprecated function: `id`"


dotsdl:
It appears that our use of `neo4j` 5.18+ yields a ton of deprecation warnings similar to:

```
WARNING  neo4j.notifications:result.py:321 Received notification from DBMS server: {severity: WARNING} {code: Neo.ClientNotification.Statement.FeatureDeprecationWarning} {category: DEPRECATION} {title: This feature is deprecated and will be removed in future versions.} {description: The query used a deprecated function: `id`.} {position: line: 5, column: 8, offset: 130} for query: "UNWIND $data AS r\nMERGE (_:GufeTokenizable {_scoped_key:r['_scoped_key']})\nSET _ += r\nSET _:DummyProtocolB:GufeTokenizable\nRETURN id(_)"
```

The complaint comes from the use of `id(_)` at the end of the query, and likely originates from our subgraph machinery in `alchemiscale.storage.subgraph`.

Is there an alternative to using this function that still gives us what we need for `SubGraph` behavior?

PR #300, opened 2024-09-04
Bumping CI Neo4j version to 5.19

ianmkenney:
grolt or py2neo fails with neo4j >= 5.19

### DIFF
--- a/alchemiscale/tests/integration/conftest.py
+++ b/alchemiscale/tests/integration/conftest.py
@@ -103,7 +103,7 @@ class TestProfile:
 
 # TODO: test with full certificates
 neo4j_deployment_profiles = [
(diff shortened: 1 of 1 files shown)


dotsdl:
We should wait to merge any solution to #291 until after we have resolved this, otherwise it's possible that our changes pass our tests for 5.18 but don't actually work for e.g. 5.22.

dotsdl:
Blocking #292.

dotsdl:
@ianmkenney can we characterize these failure further? Do we know why we see test suite failures when we set neo4j > 5.18?

ianmkenney:
It's not clear to me why the failures are happening right now. I agree we need to figure these out fairly soon.

ianmkenney:
This was fixed with https://github.com/datryllic/grolt/pull/1

Issue #304, opened 2024-09-17
Replace "passlib" with direct calls to bcrypt


LilDojd:
## Description:

It has come to my attention that `passlib` is no longer actively maintained, with the last release dating back to 2020. This raises concerns about potential CVE and long-term compatibility.

Furthermore, when using `bcrypt` versions higher than 4.0.1, I encountered an issue similar to https://github.com/pyca/bcrypt/issues/684. This suggests that `passlib` may not be compatible with the latest versions of `bcrypt` that is installed with conda in your environments.

Pinning `bcrypt` to version 4.0.1 is not a sustainable solution, as it could expose users to future security vulnerabilities that are addressed in newer releases. 

## Proposal:

I recommend replacing the usage of `passlib` with the bcrypt library directly. I will draft a PR shortly. This is not a high priority issue, so feel free to triage as you please

## References:

- [Passlib release history](https://passlib.readthedocs.io/en/stable/history/)

dotsdl:
Thanks for raising this @LilDojd! Really appreciate your help with this. :pray: 

Will review your PR #306 as part of our next sprint cycle.

PR #306, opened 2024-09-18
Replace passlib with bcrypt

LilDojd:
## Summary:

This pull request replaces the usage of the `passlib` library with `bcrypt` for password hashing and verification. 

## Changes:

- remove `passlib` from conda environments and auth.py
- Added `BcryptPasswordHandler` with same defaults as in `passlib.context.CryptContext`

See #304 for motivation

### DIFF
--- a/alchemiscale/security/auth.py
+++ b/alchemiscale/security/auth.py
@@ -4,21 +4,73 @@
 
 """
 
-from datetime import datetime, timedelta
-from typing import Union, Optional
 import secrets
+from datetime import datetime, timedelta
+from typing import Optional, Union
 
+import bcrypt
 from fastapi import HTTPException, status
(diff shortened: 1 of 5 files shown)


dotsdl:
@LilDojd thanks again for doing this!

I simplified the implementation to avoid unnecessary state in the `BcryptPasswordHandler`, along with comments on the rationale for various bits. Let me know what you think of these changes. 

We want any changes made here to allow existing `alchemiscale` deployments with existing identities/keys to still work, and I *think* this accomplishes that! Please do take a skeptical eye to it though. If it looks good to you, I think we're good to merge.

dotsdl (approved):
Great work @LilDojd! Really do appreciate this contribution, and happy we can stop seeing those `bcrypt` errors out of `passlib`!

PR #310, opened 2024-09-30
[DNM] Test if openmm 8.2 beta works

mikemhenry:


mikemhenry:
@dotsdl just wanted to check the openmm 8.2 beta didn't cause us any issues and it looks good!

PR #316, opened 2024-10-22
Switch to matching against `SettingsBaseModel` instead of `Settings` in `gufe_to_subgraph`

dotsdl:


dotsdl:
@ianmkenney this appears to work! And with the way we do the reverse operation in `Neo4jStore._node_to_gufe`, this change doesn't impact deserialization at all.

ianmkenney (approved):
This LGTM. My only concern would be the direct use of a `SettingsBaseModel` for your settings since it _seems_ like it should be an abstract class (although the limited information added in the `Settings` derived class doesn't really support that notion). In the end, if a users `Task` fails because of poorly defined settings, it doesn't technically break anything in `alchemiscale`.

dotsdl:
Thanks @ianmkenney! Merging!

PR #317, opened 2024-10-25
Addressing failures in CI

ianmkenney:
It's not exactly clear what the main cause is, but CI is failing on [main](https://github.com/OpenFreeEnergy/alchemiscale/actions/runs/11196799670/job/32072143760) (previously passing) and other [PRs](https://github.com/OpenFreeEnergy/alchemiscale/actions/runs/11507970786/job/32035228211).

The main issue seems to be that a `fastapi` dependency, `python-multipart`, isn't installing correctly.

### DIFF
--- a/.github/workflows/ci-integration.yml
+++ b/.github/workflows/ci-integration.yml
@@ -32,7 +32,7 @@ jobs:
           - "3.12"
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
 
       - uses: mamba-org/setup-micromamba@v1
         with:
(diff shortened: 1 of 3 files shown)


ianmkenney:
The difference in the breaks seems to be `python-multipart=0.0.12` -> `python-multipart=0.0.14`. According to the [package history](https://pypi.org/project/python-multipart/#history) 0.0.13 and 0.0.14 have been yanked.

dotsdl:
Discussion for `python-multipart` here: https://github.com/Kludex/python-multipart/issues/170

Agree with pinning for now.

PR #324, opened 2024-11-07
Closes #297

### LINKED ISSUE #297: Passing in malformed inputs as "ScopedKeys" gives raw 500 error
Passing in a malformed key to one of the API endpoints (e.g `/exists/blah`) causes a 500 HTTP exception to be thrown. 

This is caused by attempting to unpack the malformed key to a ScopedKey, here: https://github.com/openforcefield/alchemiscale/blob/main/alchemiscale/interface/api.py#L99C40-L100C1

Where an exception should be caught if the key cannot be unpacked into its components. This is possibly a problem in other API functions also and probably needs a more general solution.

dotsdl:
Closes #297.

### DIFF
--- a/alchemiscale/interface/client.py
+++ b/alchemiscale/interface/client.py
@@ -88,7 +88,7 @@ class AlchemiscaleClient(AlchemiscaleBaseClient):
                 "Scope for a ScopedKey must be specific; it cannot contain wildcards."
             )
 
-    def check_exists(self, scoped_key: ScopedKey) -> bool:
+    def check_exists(self, scoped_key: Union[ScopedKey, str]) -> bool:
         """Returns ``True`` if the given ScopedKey represents an object in the database."""
         return self._get_resource(f"/exists/{scoped_key}")
 
--- a/alchemiscale/interface/client.py
+++ b/alchemiscale/interface/client.py
@@ -88,7 +88,7 @@ class AlchemiscaleClient(AlchemiscaleBaseClient):
                 "Scope for a ScopedKey must be specific; it cannot contain wildcards."
             )
 
-    def check_exists(self, scoped_key: Union[ScopedKey, str]) -> bool:
+    def check_exists(self, scoped_key: ScopedKey | str) -> bool:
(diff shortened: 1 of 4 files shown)


dotsdl:
Ready for review @ianmkenney!

alchemiscale/interface/client.py
```diff
@@ -88,7 +88,7 @@ def get_scoped_key(self, obj: GufeTokenizable, scope: Scope) -> ScopedKey:
                 "Scope for a ScopedKey must be specific; it cannot contain wildcards."
             )
 
-    def check_exists(self, scoped_key: ScopedKey) -> bool:
+    def check_exists(self, scoped_key: Union[ScopedKey, str]) -> bool:
```
ianmkenney, line 91: Now that we're on Python 3.10+, we can switch these ugly `Union`s to `|`. I know there are more in the code base and this might not be the best place to do this, but figured I'd bring it up.

```suggestion
    def check_exists(self, scoped_key: ScopedKey | str) -> bool:
```

alchemiscale/tests/integration/interface/client/test_client.py
```diff
@@ -86,10 +86,38 @@ def test_create_network(
...
+        an_sk = an_sks[0]
+        an_sk_nonexistent = ScopedKey(
+            gufe_key=GufeKey("AlchemicalNetwork-lol"), **scope_test.dict()
+        )
+        assert not user_client.check_exists(an_sk_nonexistent)
+
+        # check that we get an exception when we try a malformed key
+        with pytest.raises(
+            AlchemiscaleClientError,
+            match="Status Code 422 : Unprocessable Entity : input does not appear to be a `ScopedKey`",
```
ianmkenney, line 117: nice!

ianmkenney (approved):
Looks great!

alchemiscale/tests/integration/interface/client/test_client.py
dotsdl: :grin:

PR #325, opened 2024-11-07
Use python < 3.13 for building docs

ianmkenney:
Python v3.13 removed imghdr, which is used by the versions sphinx we use for building our docs. Since we don't officially support 3.13 yet, we can get the docs working by adding this constraint. A better option would be to use newer versions of Sphinx, but that can be done in a different commit.

See https://github.com/sphinx-doc/sphinx/issues/10440


dotsdl:
Awesome, thank you @ianmkenney! I agree with your proposed solution here.

PR #336, opened 2024-11-25
[WIP] Add use of zstd compression on compute services

### LINKED ISSUE #220: Add use of zstd compression on compute services for pushing to API service, storing at-rest, and fast decompression on retreival via clients
Currently, we perform no compression at-rest for `ProtocolDAGResult`s produced by compute services. This yields larger-than-necessary JSON files on the object store, and either larger-than-necessary transmission payloads or unnecessary compression/decompression in-transit for files that do not ever change on the object store.

Instead, we should compress once `ProtocolDAGResult`s on creation on the compute service itself, then push this compressed artifact to the compute API for delivery to the object store. Requests by clients for this artifact would then pull the compressed artifact as-is with no in-transit compression/decompression required. If using [`zstd`](https://en.wikipedia.org/wiki/Zstd) by way of e.g. [`python-zstandard`](https://python-zstandard.readthedocs.io/en/latest/concepts.html), decompression client-side can be very fast, mitigating wait times for users in retrieving results. 

Because the compression is done on the compute services themselves, this scheme puts minimal load on the API services receiving or retrieving these artifacts. Relatively slow compression happens only once on creation, and very fast decompression can happen many times over the lifetime of the artifact.

Importantly, for all cases of compressed `ProtocolDAGResult` retrieval, **compression in-transit should be disabled** to avoid wasteful double-compression/decompression.

ianmkenney:
This PR closes #220. It modifies the behavior of set_task_results in the compute client and api to use compressed keyed chain representations of ProtocolDAGResults instead of simple JSON serialization as the intermediate format.

### DIFF
--- a/alchemiscale/compute/api.py
+++ b/alchemiscale/compute/api.py
@@ -9,10 +9,12 @@ import os
 import json
 from datetime import datetime, timedelta
 import random
+import base64
 
 from fastapi import FastAPI, APIRouter, Body, Depends
 from fastapi.middleware.gzip import GZipMiddleware
-from gufe.tokenization import GufeTokenizable, JSON_HANDLER
+from gufe.tokenization import GufeTokenizable, JSON_HANDLER, KeyedChain
+import zstandard as zstd
 
 from ..base.api import (
     QueryGUFEHandler,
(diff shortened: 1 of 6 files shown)


dotsdl (approved):
Looks fantastic @ianmkenney! Thanks for adding elegant compression-at-rest for `ProtocolDAGResult` objects!

PR #337, opened 2024-11-30
`SynchronousComputeService` now properly claims tasks with protocols filter

dotsdl:
We weren't actually using the `protocols` setting for the `SynchronousComputeService`, so `Task` claims weren't being filtered if this was set to anything but `None`.


dotsdl:
@ianmkenney can I get a quick review from you on this one?

ianmkenney (approved):
lgtm

PR #359, opened 2025-02-13
Remove Python 3.10 from CI, add SPEC0 badge

### LINKED ISSUE #353: Drop Python 3.10 from support in line with SPEC 0
In line with [SPEC 0](https://scientific-python.org/specs/spec-0000/), the recommended support window for Python 3.10 ended in Q4 2024. Since we are likely to adopt features that go beyond Python 3.10 soon, it may make sense to officially drop support for this within `alchemiscale`.

Should we also adopt SPEC 0 formally, and add the [badge](https://scientific-python.org/specs/spec-0000/#badges) to the README?

dotsdl:
Closes #353.

### DIFF
--- a/.github/workflows/ci-integration.yml
+++ b/.github/workflows/ci-integration.yml
@@ -27,7 +27,6 @@ jobs:
       matrix:
         os: ['ubuntu']
         python-version:
-          - "3.10"
           - "3.11"
           - "3.12"
 
--- a/.github/workflows/ci-integration.yml
+++ b/.github/workflows/ci-integration.yml
@@ -29,6 +29,7 @@ jobs:
         python-version:
           - "3.11"
           - "3.12"
+          - "3.13"
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/ci-integration.yml
+++ b/.github/workflows/ci-integration.yml
@@ -29,7 +29,6 @@ jobs:
         python-version:
           - "3.11"
           - "3.12"
-          - "3.13"
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


.github/workflows/ci-integration.yml
```diff
@@ -27,7 +27,6 @@ jobs:
       matrix:
         os: ['ubuntu']
         python-version:
-          - "3.10"
           - "3.11"
           - "3.12"
```
ianmkenney, line 32: Do we also want to add 3.13 here?

dotsdl: We can! I believe the `openfe` stack doesn't support 3.13 yet (due to `ambertools` IIRC), so it might just fail systematically.

dotsdl:
Getting the following when trying to test against Python 3.13: https://github.com/OpenFreeEnergy/alchemiscale/actions/runs/13317902605/job/37196314556?pr=359#step:3:35

```
 error    libmamba Could not solve for environment specs
      The following packages are incompatible
      ├─ openmmforcefields >=0.14.1  is installable and it requires
      │  └─ python >=3.10,<3.13  with the potential options
      │     ├─ python [3.10.0|3.10.1|...|3.12.8], which can be installed;
      │     └─ python [3.12.0rc3|3.13.0rc1|3.13.0rc2|3.13.0rc3] would require
      │        └─ _python_rc, which does not exist (perhaps a missing channel);
      └─ python 3.13**  is not installable because there are no viable options
         ├─ python [3.12.0rc3|3.13.0rc1|3.13.0rc2|3.13.0rc3], which cannot be installed (as previously explained);
         └─ python [3.13.0|3.13.1] conflicts with any installable versions previously reported.
  critical libmamba Could not solve for environment specs
```

hmacdope:
Trying 3.13 for openmmff here: https://github.com/openmm/openmmforcefields/pull/362

IAlibay:
@dotsdl unfortunately no py3.13 for openfe for a while to go. We're blocked by a couple of upstream dependencies, one of which I'm in charge of but haven't been able to find the time to deal with it.

dotsdl:
Thanks @hmacdope and @IAlibay, I figured this was the case but figured we'd try it out. Since it's known that we won't be able to support Python 3.13 for some time I'll remove it from our CI testing here.

ianmkenney (approved):

Issue #266, opened 2024-04-10
Decrease Docker image size


dotsdl:
As of this writing, the `alchemiscale-server` image is 6.23GB in size. The `alchemiscale-compute` image is similarly large. We should identify opportunities to reduce these, in particular any very large dependencies that are being pulled in via `conda`.

mikemhenry:
I've looked into this a bit, right now our conda env is ~5.2GB with `cudatoolkit` the largest:

![Image](https://github.com/user-attachments/assets/02ab07e7-6988-4658-aa97-d605a99f124f)

Have you done any testing to see if we can use `cuda >=12` on different hpc resources? Newer versions of cuda split the toolkit into different components so we can save some space. 

We can save ~100 MB doing a multistage build, but at this point most of the space savings are going to be in slimming down our conda env to have just the runtime dependencies needed.

mikemhenry:
Something to consider:

Users will need to use the cuda version that matches their execution environment, so we build a matrix of different cuda versions (including no cuda) for images so that users don't need to build the images themselves.

dotsdl:
Looks like no matter what we do, we end up pulling in `cuda` or `cudatoolkit` in some form. It appears [`gufe` is dependent on `openmm`](https://github.com/conda-forge/gufe-feedstock/blob/main/recipe/meta.yaml), and installing `openmm` appears to pull in `cuda` on its own unprompted:

```
! micromamba create -n openmm-test -c conda-forge openmm
conda-forge/linux-64                                        Using cache
conda-forge/noarch                                          Using cache

error    libmamba Could not lock non-existing path '/home/david/.mamba/pkgs'

Transaction

  Prefix: /home/david/micromamba/envs/openmm-test

  Updating specs:

   - openmm


  Package                       Version  Build                 Channel           Size
───────────────────────────────────────────────────────────────────────────────────────
  Install:
───────────────────────────────────────────────────────────────────────────────────────

  + _libgcc_mutex                   0.1  conda_forge           conda-forge     Cached
  + _openmp_mutex                   4.5  2_gnu                 conda-forge     Cached
  + bzip2                         1.0.8  h4bc722e_7            conda-forge     Cached
  + ca-certificates           2025.1.31  hbcca054_0            conda-forge     Cached
  + cudatoolkit                  11.8.0  h4ba93d1_13           conda-forge     Cached
  + kernel-headers_linux-64      3.10.0  he073ed8_18           conda-forge     Cached
[...]

PR #360, opened 2025-02-14
Updated `gufe` and `openfe` versions in conda envs, added `cuda-version > 12` to avoid pulling `cudatoolkit`

dotsdl:
Attempting to see if this decreases image sizes; this would also be our approach for building 0.6.0 images.

Addresses #266.

### DIFF
--- a/devtools/conda-envs/alchemiscale-client.yml
+++ b/devtools/conda-envs/alchemiscale-client.yml
@@ -9,7 +9,7 @@ dependencies:
 
   # alchemiscale dependencies
   - gufe=1.2.0
-  - openfe=1.3.0
+  - openfe=1.2.0
   - requests
   - click
   - httpx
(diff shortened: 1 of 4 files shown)


dotsdl:
With these changes, [`alchemiscale-server` image is now down to ~3.6GB](https://github.com/OpenFreeEnergy/alchemiscale/actions/runs/13321976305/job/37208144482#step:5:1264), and [`alchemiscale-compute` image is down to ~3.4GB](https://github.com/OpenFreeEnergy/alchemiscale/actions/runs/13321976305/job/37208144671#step:5:1147). This is nearly half the size these images were previously!

Can I get your review on this @mikemhenry?

mikemhenry:
Sweet! And we can likely slim them down even more with removing some of the -dev libraries that get pulled in. The only issue is if people need older CUDA versions for the images, but I don't think we are really on the hook since someone can build their own image with a different CUDA version if needed

mikemhenry (approved):
LGTM

PR #367, opened 2025-02-20
Fix sporadic test failures

ianmkenney:
This PR addresses two classes of errors we see in our tests:

1. A logic error within the tests where we assert certain transformations exist when we randomly deleted them. (local/CI)
2. An error that seems to be caused by a race condition. We haven't been able to reproduce this locally.

The first error seems to be fixed by simply sorting transformations before pruning. The second we might be able to safely ignore (xfail if the error has the right message). I'm not opposed to reverting this second change since it might be a little aggressive.

### DIFF
--- a/alchemiscale/tests/integration/interface/client/test_client.py
+++ b/alchemiscale/tests/integration/interface/client/test_client.py
@@ -2441,7 +2441,24 @@ class TestClient:
 
         # execute the actioned tasks and push results directly using statestore and object store
         with tmpdir.as_cwd():
-            protocoldagresults = self._execute_tasks(actioned_tasks, n4js, s3os_server)
+            try:
+                protocoldagresults = self._execute_tasks(
+                    actioned_tasks, n4js, s3os_server
+                )
+            # in CI workflows the above operation seems to randomly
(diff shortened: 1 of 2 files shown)


alchemiscale/tests/integration/interface/client/test_client.py
```diff
@@ -2441,7 +2441,24 @@ def test_get_task_failures(
...
         # execute the actioned tasks and push results directly using statestore and object store
         with tmpdir.as_cwd():
-            protocoldagresults = self._execute_tasks(actioned_tasks, n4js, s3os_server)
+            try:
+                protocoldagresults = self._execute_tasks(
+                    actioned_tasks, n4js, s3os_server
+                )
+            # in CI workflows the above operation seems to randomly
+            # fail with what we believe is a race condition due to the
+            # workers limited hardwear. We are unable to reporoduce
```
dotsdl, line 2450:
```suggestion
            # worker's limited resources. We are unable to reproduce
```

dotsdl (approved):
This looks great @ianmkenney! Thank you for doing this!

One small edit to docstring; I'll go ahead and accept this and merge!

PR #370, opened 2025-02-27
Fix broken codepath in compute API for resolving task restarts for failed ProtocolDAGResults

dotsdl:
Also adds a test for the above.

## Checklist
* [ ] Added a ``news`` entry

## Developers certificate of origin
- [ ] I certify that this contribution is covered by the MIT License [here](https://github.com/OpenFreeEnergy/openfe/blob/main/LICENSE) and the **Developer Certificate of Origin** at <https://developercertificate.org/>.

### DIFF
--- a/alchemiscale/compute/api.py
+++ b/alchemiscale/compute/api.py
@@ -365,7 +365,7 @@ async def set_task_result(
             pdr.protocol_unit_failures, result_sk
         )
         n4js.set_task_error(tasks=[task_sk])
-        n4js.resolve_task_restarts(tasks=[task_sk])
+        n4js.resolve_task_restarts(task_scoped_keys=[task_sk])
 
     return result_sk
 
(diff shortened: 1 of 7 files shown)


dotsdl:
@ianmkenney can I get your review on this? This is a bug that I discovered in working on `alchemiscale-fah` release following `alchemiscale` 0.6.0. We'll want to merge and release 0.6.1 as soon as possible, since this will currently impact all failed `ProtocolDAGResult`s on deployed instances.

alchemiscale/compute/api.py
```diff
@@ -365,7 +365,7 @@ async def set_task_result(
             pdr.protocol_unit_failures, result_sk
         )
         n4js.set_task_error(tasks=[task_sk])
-        n4js.resolve_task_restarts(tasks=[task_sk])
+        n4js.resolve_task_restarts(task_scoped_keys=[task_sk])
```
ianmkenney, line 368: oof good catch

ianmkenney: This also appears to be the only place we use `task_scoped_keys` so explicitly, we could remove it.

ianmkenney (approved):
good catch! lgtm

PR #379, opened 2025-03-07
switch to using setuptools_scm

### LINKED ISSUE #378: Switch to using setuptools_scm
`setuptools_scm` I find works better than `versioningit` -- I like that it supports using an environmental variable to set the version, which means that when building a package on conda-forge, you don't need to clone the whole git repo but instead can build the package off a tarball and still get the correct version. 

I am happy to make the PR that switches to it -- looking at the codebase this will be an easy swap in the `pyproject.toml` and next release I can update the feedstock to work with it.

mikemhenry:
I noticed we use some python 3.10 syntax, so really 3.10 is the minium python version.

Fixes #378


dotsdl (approved):
Oh very cool! Is it the case that `versioningit` is just no longer necessary? If this does what we want it to I'm very much for the simplification!

dotsdl:
@mikemhenry ah, you answered my question in #378! If we're confident this should work well, feel free to merge!

mikemhenry:
Yes! We don't need it anymore, so on next release we can update the conda-forge feedstock to remove it as a host dep + also not clone the repo to build the package

dotsdl:
Awesome, thanks so much @mikemhenry! I love it when we can simplify things!

PR #381, opened 2025-03-07
Fix docs

ianmkenney:
Currently the API reference is silently failing to build and it seems to be because we don't have `diskcache` in the `docs.yml` environment recipe.

Edit: I don't fully understand why this is a problem now when it should fail with other packages like `gufe` in the same way.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,7 +1,7 @@
 version: 2
 
 build:
-  os: "ubuntu-20.04"
+  os: "ubuntu-24.04"
   tools:
     python: "mambaforge-4.10"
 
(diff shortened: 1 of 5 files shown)


ianmkenney:
There are some more warnings that need to be handled, but at the moment the live docs don't have an API reference. This fixes the initial issue and incorporates some of the changes introduced in #372 (which can be closed after merging this). I'll follow up with another PR trying to make warnings trigger a failure.

dotsdl (approved):
Looks great @ianmkenney! Thanks for doing this!

dotsdl:
@ianmkenney please merge if satisfied!

mikemhenry (approved):
Weird that `Optional` works but ` | None` doesn't!

PR #390, opened 2025-03-18
Prevent warnings when environment variable is not set, and parameter is supplied

LilDojd:
I have just realized, that the type checker was right to [question](https://github.com/OpenFreeEnergy/alchemiscale/pull/368#discussion_r1980625297) my match statement in the client parameter. The third case did actually match `(Some(parameter), None)` which led to warnings spam.

This small PR makes the `case` statement exhaustive and adds exhaustive tests for all combinations of env/python variables


LilDojd:
@dotsdl Hi, I think this is ready, could you please assign reviewer?

ianmkenney:
Thanks for the quick fix @LilDojd!

PR #395, opened 2025-04-07
[WIP] Optimize submission of AlchemicalNetwork objects to the statestore

### LINKED ISSUE #375: Further `KeyedChain`-based optimization of `AlchemicalNetwork` submission
Previously, we greatly reduced submission times and memory use for large `AlchemicalNetwork`s via the `AlchemiscaleClient` by introducing the concept of `KeyedChain`s, which removed the repetition of components that plagued the `dict` representation of an `AlchemicalNetwork`. However, even with this optimization, submitting `AlchemicalNetwork`s with thousands of `ChemicalSystem`s and many more `Transformation`s remains slow, taking 40 minutes or more in some cases.

To address this, we should attempt to take advantage of the `KeyedChain` representation further, skipping if possible going from `KeyedChain` to `GufeTokenizable` and then to `py2neo.Subgraph` before translating this to Cypher queries for injection into Neo4j. Given that all the information we require for the Neo4j representation is already in the `KeyedChain` representation in a reasonable form, it should be possible to proceed from a `KeyedChain` into a series of Cypher queries, building up the required nodes and relationships in `KeyedChain` order.

If this works, it would have the side-effect of removing our dependency of the now-dead `py2neo` project, solving #144.

ianmkenney:
closes #375

### DIFF
--- a/alchemiscale/storage/statestore.py
+++ b/alchemiscale/storage/statestore.py
@@ -260,8 +260,8 @@ class Neo4jStore(AlchemiscaleStateStore):
         node["_scoped_key"] = str(scoped_key)
 
         for key, value in sdct.items():
-            match (key, value):
-                case (_, dict()):
+            match value:
+                case dict():
                     if all([isinstance(x, GufeTokenizable) for x in value.values()]):
                         for k, v in value.items():
                             node_ = subgraph_ = self.gufe_nodes.get(
@@ -293,7 +293,7 @@ class Neo4jStore(AlchemiscaleStateStore):
                     else:
                         node[key] = json.dumps(value, cls=JSON_HANDLER.encoder)
                         node["_json_props"].append(key)
-                case (_, list()):
+                case list():
                     # lists can only be made of a single, primitive data type
                     # we encode these as strings with a special starting indicator
                     if isinstance(value[0], (int, float, str)) and all(
@@ -331,7 +331,7 @@ class Neo4jStore(AlchemiscaleStateStore):
                     else:
                         node[key] = json.dumps(value, cls=JSON_HANDLER.encoder)
                         node["_json_props"].append(key)
-                case (_, tuple()):
+                case tuple():
                     # lists can only be made of a single, primitive data type
                     # we encode these as strings with a special starting indicator
                     if not (
@@ -340,10 +340,10 @@ class Neo4jStore(AlchemiscaleStateStore):
                     ):
                         node[key] = json.dumps(value, cls=JSON_HANDLER.encoder)
                         node["_json_props"].append(key)
-                case (_, SettingsBaseModel()):
+                case SettingsBaseModel():
                     node[key] = json.dumps(value, cls=JSON_HANDLER.encoder, sort_keys=True)
                     node["_json_props"].append(key)
-                case (_, GufeTokenizable()):
+                case GufeTokenizable():
                     node_ = subgraph_ = self.gufe_nodes.get(
                         (value.key, scope.org, scope.campaign, scope.project)
                     )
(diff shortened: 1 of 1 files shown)


alchemiscale/storage/statestore.py
```diff
@@ -242,138 +242,139 @@ def reset(self):
...
+                node["_json_props"].append(key)
 
-        subgraph = subgraph | node
+        def handle_tuple(node, key, values):
+            if not (
+                isinstance(values[0], (int, float, str))
+                and all((isinstance(x, type(values[0])) for x in values))
+            ):
+                node[key] = json.dumps(values, cls=JSON_HANDLER.encoder)
+                node["_json_props"].append(key)
```
dotsdl, line 334: I realize now that this captures the previous behavior, but it looks like we weren't handling the case where this `if` statement gives `False`. In that case, the `tuple` will be lost entirely.

Do you think this is a legit bug @ianmkenney?

alchemiscale/storage/statestore.py
```diff
@@ -242,138 +242,139 @@ def reset(self):
...
+            if all(map(is_gufe_dict, dct.values())):
+                for k, v in dct.items():
+                    rel_key = v[":gufe-key:"]
+                    update_relationships(
+                        node, get_previous_node(rel_key), attribute=key, key=k
+                    )
             else:
-                node[key] = value
+                node[key] = json.dumps(dct, cls=JSON_HANDLER.encoder)
+                node["_json_props"].append(key)
```
dotsdl, line 326: Nitpick: can we put these methods in the same order as they appear in `process_keyed_dict`?

alchemiscale/storage/statestore.py
```diff
@@ -242,138 +242,139 @@ def reset(self):
...
+                        handle_dict(node, key, value)
+                    case list():
+                        handle_list(node, key, value)
+                    case tuple():
+                        handle_tuple(node, key, value)
+                    case SettingsBaseModel():
+                        node[key] = json.dumps(
+                            value, cls=JSON_HANDLER.encoder, sort_keys=True
+                        )
+                        node["_json_props"].append(key)
```
dotsdl, line 355: Perhaps put this in its own function for consistency with the others?

alchemiscale/storage/statestore.py
dotsdl: We should probably just JSONify all `tuple`s to distinguish them from the way `list`s are handled in `_node_to_gufe`.

dotsdl (approved):
Looks great @ianmkenney! Can you review #400, and after merge there, add a news entry for this and address the comments?

After the above, please merge if you're satisfied!

alchemiscale/storage/statestore.py
ianmkenney: Definitely a bug in both implementations. I forgot to drop a todo in there to get back to it. I'll look into a solution for it tomorrow.

alchemiscale/storage/statestore.py
ianmkenney: sure

alchemiscale/storage/statestore.py
ianmkenney: sure

alchemiscale/storage/statestore.py
```diff
@@ -242,138 +242,139 @@ def reset(self):
...
-                node["_json_props"].append(key)
-            elif isinstance(value, GufeTokenizable):
-                node_ = subgraph_ = self.gufe_nodes.get(
-                    (value.key, scope.org, scope.campaign, scope.project)
+        Parameters
+        ----------
+        keyed_chain: KeyedChain
+            The keyed chain to convert into a subgraph.
+        scope: Scope
+            The scope to assign to the Subgraph Node objects.
```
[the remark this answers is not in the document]
ianmkenney: will do

dotsdl (approved):
Looks great @ianmkenney!

Please add a news entry, then merge if satisfied!

PR #400, opened 2025-04-14
Add CHANGELOG via rever

dotsdl:
Since we manage a human-readable CHANGELOG in `gufe` via `rever` and it works pretty well, I'm proposing we adopt the practice here.


ianmkenney (approved):
I've not used rever personally, but from what I can see this is what's needed.

Issue #404, opened 2025-04-21
Deprecated `utcnow` function in Python>=3.12


ianmkenney:
As of [Python 3.12](https://docs.python.org/3/whatsnew/3.12.html#deprecated), the `utcnow` function has been deprecated. It is recommended to use the following as a replacement: `datetime.datetime.now(tz=datetime.UTC)`.

`datetime.UTC` was added in [Python 3.11](https://docs.python.org/3/whatsnew/3.11.html#datetime) and is compatible with the currently supported versions of Python for `alchemiscale`.

PR #418, opened 2025-06-25
update for `utcnow` deprecation

atravitz:
resolves https://github.com/OpenFreeEnergy/alchemiscale/issues/404

it looks like  `datetime.datetime.now(tz=datetime.UTC)` adds timezone information that breaks some string conversion on the js side.  using `datetime.datetime.now(tz=None)` resolves that, but I think I'd prefer to find a solution that passes `datetime.UTC` to be more explicit.

### DIFF
--- a/alchemiscale/compute/api.py
+++ b/alchemiscale/compute/api.py
@@ -104,7 +104,7 @@ def register_computeservice(
     compute_service_id,
     n4js: Neo4jStore = Depends(get_n4js_depends),
 ):
-    now = datetime.datetime.now(tz=datetime.UTC)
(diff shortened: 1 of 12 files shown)


dotsdl:
Thanks @atravitz! We definitely want to ensure our times are in UTC to avoid local times being reported between services that are geographically separated. Can you set these changes to use `datetime.datetime.now(tz=datetime.UTC)` and we can then work out how to fix the breakages?

dotsdl:
Hey @atravitz, were you able to loop back onto this one? No rush if not!

PR #430, opened 2025-08-21
Update test env to latest deps

dotsdl:
Includes:
- `gufe` v1.6.0
- `openfe` v1.6.0
- `openmm` v8.2.0

We want to check that we also get in latest `openff-toolkit` and `openff-interchange`, as well as `nagl` and `pytorch` as dependencies.

### DIFF
--- a/devtools/conda-envs/test.yml
+++ b/devtools/conda-envs/test.yml
@@ -7,8 +7,8 @@ dependencies:
   - cuda-version >=12
 
   # alchemiscale dependencies
-  - gufe=1.3.0
-  - openfe=1.4.0
+  - gufe=1.6.0
+  - openfe=1.6.0
   - pydantic >2
   - pydantic-settings
   - async-lru
@@ -52,9 +52,8 @@ dependencies:
   - moto
 
   # additional pins
-  - openmm=8.1.2
+  - openmm=8.2.0
   - openmmforcefields>=0.14.2
-  - openff-units=0.2.2
 
   - pip:
     - git+https://github.com/datryllic/grolt # neo4j test server deployment


devtools/conda-envs/test.yml
```diff
@@ -52,9 +52,8 @@ dependencies:
   - moto
 
   # additional pins
-  - openmm=8.1.2
+  - openmm=8.2.0
   - openmmforcefields>=0.14.2
-  - openff-units=0.2.2
```
IAlibay, line 57: This should now be at openff-units 0.3 (openfe feedstock should have pinned it).

devtools/conda-envs/test.yml
```diff
@@ -7,8 +7,8 @@ dependencies:
   - cuda-version >=12
 
   # alchemiscale dependencies
-  - gufe=1.3.0
-  - openfe=1.4.0
+  - gufe=1.6.0
+  - openfe=1.6.0
```
IAlibay, line 11: Not necessary for this environment, but just an FYI that we'll need openff-toolkit pinned at 0.17.0 in the runtime environment.

IAlibay (commented):
Just a couple of comments (will provide more context on slack).

devtools/conda-envs/test.yml
dotsdl: Oh got it; do you recommend we pin here?

devtools/conda-envs/test.yml
dotsdl: Thank you! Yes, this is just the test env. I'll be pinning explicitly in our `alchemiscale.org-deployment` envs.

devtools/conda-envs/test.yml
IAlibay: Probably not, we're pretty good at strictly pinning it in the openfe and gufe feedstocks.

PR #433, opened 2025-09-16
Allow server-side copying and merging of AlchemicalNetworks

### LINKED ISSUE #221: Add `AlchemiscaleClient.copy_network`, `.merge_networks`, and `.merge_scopes` methods
Currently users create `AlchemicalNetwork`s in separate `Scope`s if they want to independently generate results for each, or create `AlchemicalNetwork`s in the same `Scope` if they would like results to be shared for common `Transformation`s. This requires *a priori* planning for how to make use of results between `AlchemicalNetwork`s; what about when a user would like to take existing results from `AlchemicalNetwork`s in different `Scope`s and combine them into another `Scope`?

To enable this, we propose exposing the following methods on the `AlchemiscaleClient`:
- `copy_network(<network_scoped_key>, <scope>)` : copies an entire `AlchemicalNetwork` and all existing `Task`s and associated `ProtocolDAGResultRef`s to the given `Scope`, which may or may not already feature other `AlchemicalNetwork`s with the same `Transformation`s
- `merge_networks(<network_scoped_key>, <network_scoped_key>, ..., <scope>, name=None)` : creates a new `AlchemicalNetwork` in the given `Scope` from the `ChemicalSystem`s and `Transformation`s of any number of other `AlchemicalNetwork`s, optionally with a `name`.
- `merge_scopes(<scope>, <scope>, ..., <scope>)` : copies all `AlchemicalNetwork`s in any number of given `Scope`s into another `Scope`, similarly to `copy_network`

These methods will create new nodes and relationships in the state store, but shouldn't need to touch the object store, requiring no duplication of large objects there.

ianmkenney:
closes #221


alchemiscale/storage/statestore.py
```diff
@@ -876,6 +883,227 @@ def delete_network(
...
+                    subchain = KeyedChain(network_keyed_chain[: index + 1])
+                    transformation = kc_to_gufe(subchain, gts)
+                    subchain_cache[transformation] = subchain
+                    try:
+                        idx = transformation_data.index(transformation)
+                        data = transformation_data[idx]
+                    except ValueError:
+                        data = TransformationData(transformation)
+                        transformation_data.append(data)
+                    data.add_known_scoped_key(database_key, network_scope)
```
dotsdl: Might be able to now use `GufeTokenizable.decode_subchains` now that OpenFreeEnergy/gufe#629 is closed?

PR #449, opened 2025-11-04
feat: parallel test neo4j containers

jandom:
Each test startup is still kind of slow but at least they can be run in parallel

### DIFF
--- a/alchemiscale/compute/service.py
+++ b/alchemiscale/compute/service.py
@@ -353,6 +353,8 @@ class SynchronousComputeService:
             If `None`, the service will have no time limit.
 
         """
+        self._stop = False
+
         # add ComputeServiceRegistration
         self.logger.info("Starting up service '%s'", self.name)
         self._register()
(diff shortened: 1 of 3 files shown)


jandom:
Okay, I think I'm getting somewhere ... It was all about making sure that multiple worker don't all talk to the services 

  1. ✅ Neo4j containers: Each worker gets its own isolated container with unique ports
  2. ✅ Uvicorn API servers: Dynamic port assignment per worker (8000 + offset)
  3. ✅ S3 mock servers: Dynamic port assignment per worker (5000 + offset)

So things are no longer crashing/hanging. However, there is still no perf improvement 

```
python -m pytest -n 4 -vv  alchemiscale/tests/integration/compute/client/test_compute_client.py
============================================================== 14 passed, 32 warnings in 269.34s (0:04:29) ===============================================================

python -m pytest -n 2 -vv  alchemiscale/tests/integration/compute/client/test_compute_client.py
============================================================== 14 passed, 18 warnings in 255.86s (0:04:15) ===============================================================

python -m pytest -n 1 -vv  alchemiscale/tests/integration/compute/client/test_compute_client.py
================================================ 14 passed, 11 warnings in 306.52s (0:05:06) ================================================
```

PR #467, opened 2025-12-10
Adds ComputeServiceID form enforcement

dotsdl:
## Summary

Enforces a structured form on `ComputeServiceID` (parity with the existing `ComputeManagerID` enforcement) and refactors both into a shared `ComputeIDBase` to remove duplication.

## Changes

- **New `ComputeIDBase` class** in `alchemiscale/storage/models.py` consolidating the validation, `to_dict`/`from_dict`, `name`/`uuid` properties, and `new_from_name` constructor previously living only on `ComputeManagerID`.
- **`ComputeServiceID` and `ComputeManagerID`** now both subclass `ComputeIDBase`, so both IDs are validated identically as `{name}-{uuid}`.
- **Looser name grammar** (per discussion with @ianmkenney): names now match `^[a-zA-Z][a-zA-Z0-9_\.\:]*$` — alphabetical first character, followed by alphanumerics, underscores, periods, or colons. This replaces the previous `isalnum()` check on `ComputeManagerID`.
- **`:` as an additional delimiter** inside the name portion lets us split the ID on `-` into exactly two parts (`name`, `uuid`), rather than six fields as before. The UUID portion is now expected in hex form (no dashes).
- **Call sites updated** to use `ComputeServiceID.new_from_name(...)` instead of manually formatting `f"{name}-{uuid4()}"`:
  - `alchemiscale/compute/service.py`
  - `alchemiscale/tests/integration/conftest.py` (also switches the test name `compute-service-123` → `compute.service.123` to exercise the new delimiter rules)
- **Drive-by fix** in `alchemiscale/storage/statestore.py`: removes a stray `|\*$` alternation in `_validate_protocols` that allowed a bare `*` to slip past validation.

## Notes for reviewers

- Existing persisted `ComputeManagerID` values used dashed UUID form (6 dash-separated fields). The new parser expects hex UUIDs (no dashes) because `-` is now the single name/uuid delimiter. Worth confirming whether any stored IDs need to be migrated.
- The `_allowed` regex is defined on the base class; if `ComputeServiceID` and `ComputeManagerID` ever need to diverge on allowed characters, override there.

## Test plan

- [x] Existing unit and integration tests for `ComputeManagerID` still pass
- [x] New/updated tests cover `ComputeServiceID` validation (good and bad names, good and bad UUIDs)
- [x] `compute_service_id` integration fixture exercises the `.` delimiter case
- [ ] Verify no production callers construct `ComputeServiceID` from a legacy `{name}-{uuid-with-dashes}` string


dotsdl (commented):
I'm satisfied with this! @ianmkenney can you review?

ianmkenney (approved):
This looks reasonable. If I were to nitpick, it would be about the `_allowed` attribute name. Since it's only applied to the name and not the name + uuid, `_allowed_name_pattern` might be more appropriate.

Not familiar with the changes to the testing, but will leave that to you.

dotsdl:
Awesome, thank you @ianmkenney! Taking your suggestion!

PR #477, opened 2026-01-09
Address issue #222

### LINKED ISSUE #222: Add `scopes-exclude` config option for compute services
Currently, it is possible to specify which `Scope`s a compute service should draw `Task`s from via the `scopes` configuration key. However, in some cases it is complementary to also be able to specify `Scope`s that should be excluded. This can happen when one gives a non-specific `Scope` in `scopes`, but would like to leave out selection of `Task`s from a specific `Scope` that is a subset of that non-specific `Scope`.

We should add a `scopes-exclude` configuration option that applies after `Scope`s given in `scopes` as a filter on matches.

dotsdl:
This commit introduces a new `scopes_exclude` configuration parameter for compute services, addressing issue #222.

Changes:
- Added `scopes_exclude` field to ComputeServiceSettings with validation
- Updated ComputeService to initialize and pass scopes_exclude to client
- Updated ComputeClient.claim_tasks() to accept and transmit scopes_exclude
- Updated compute API /claim endpoint to filter taskhubs based on excluded scopes
- Filtering uses Scope.is_superset() to exclude taskhubs matching excluded scopes

The scopes_exclude option is applied after scopes as a filter, allowing users to specify broad scopes while excluding specific subset scopes from task claiming.

Fixes #222

### DIFF
--- a/alchemiscale/compute/api.py
+++ b/alchemiscale/compute/api.py
@@ -218,6 +218,7 @@ def claim_taskhub_tasks(
 @router.post("/claim")
 def claim_tasks(
     scopes: list[Scope] = Body(),
+    scopes_exclude: list[Scope] | None = Body(None, embed=True),
     compute_service_id: str = Body(),
     count: int = Body(),
     protocols: list[str] | None = Body(None, embed=True),
@@ -255,6 +256,20 @@ def claim_tasks(
     # loop might be more removable in the future with a Union like operator on scopes
     for single_query_scope in set(query_scopes):
         taskhubs.update(n4js.query_taskhubs(scope=single_query_scope, return_gufe=True))
+
+    # filter out taskhubs that match excluded scopes
+    if scopes_exclude:
+        taskhubs_filtered = {}
+        for taskhub_sk, taskhub in taskhubs.items():
+            exclude = False
+            for exclude_scope in scopes_exclude:
+                # if the exclude scope is a superset of the taskhub scope, exclude it
+                if exclude_scope.is_superset(taskhub_sk.scope):
+                    exclude = True
+                    break
+            if not exclude:
+                taskhubs_filtered[taskhub_sk] = taskhub
+        taskhubs = taskhubs_filtered
 
     # list of tasks to return
     tasks = []
--- a/alchemiscale/compute/api.py
+++ b/alchemiscale/compute/api.py
@@ -258,17 +258,12 @@ def claim_tasks(
 
     # filter out taskhubs that match excluded scopes
     if scopes_exclude:
-        taskhubs_filtered = {}
-        for taskhub_sk, taskhub in taskhubs.items():
-            exclude = False
-            for exclude_scope in scopes_exclude:
-                # if the exclude scope is a superset of the taskhub scope, exclude it
-                if exclude_scope.is_superset(taskhub_sk.scope):
-                    exclude = True
-                    break
-            if not exclude:
-                taskhubs_filtered[taskhub_sk] = taskhub
-        taskhubs = taskhubs_filtered
+        scopes_exclude_reduced = minimize_scope_space(scopes_exclude)
+        taskhubs = {
+            sk: th
+            for sk, th in taskhubs.items()
+            if not any(ex.is_superset(sk.scope) for ex in scopes_exclude_reduced)
+        }
 
     # list of tasks to return
     tasks = []
--- a/alchemiscale/compute/api.py
+++ b/alchemiscale/compute/api.py
@@ -256,20 +256,24 @@ def claim_tasks(
     for single_query_scope in set(query_scopes):
         taskhubs.update(n4js.query_taskhubs(scope=single_query_scope, return_gufe=True))
 
-    # filter out taskhubs that match excluded scopes
+    # filter out taskhubs whose scope is covered by any excluded scope
     if scopes_exclude:
         scopes_exclude_reduced = minimize_scope_space(scopes_exclude)
-        taskhubs = {
-            sk: th
-            for sk, th in taskhubs.items()
-            if not any(ex.is_superset(sk.scope) for ex in scopes_exclude_reduced)
-        }
+
+        def is_excluded(taskhub_scope: Scope) -> bool:
+            return any(
+                excluded.is_superset(taskhub_scope)
+                for excluded in scopes_exclude_reduced
+            )
+
+        taskhubs = {sk: th for sk, th in taskhubs.items() if not is_excluded(sk.scope)}
 
     # list of tasks to return
     tasks = []
 
     if len(taskhubs) == 0:
-        return []
+        # match the pad-with-``None`` behavior of the normal return path below
+        return [None] * count
 
     # claim tasks from taskhubs based on weight; keep going till we hit our
     # total desired task count, or we run out of taskhubs to draw from
--- a/alchemiscale/compute/service.py
+++ b/alchemiscale/compute/service.py
@@ -97,6 +97,11 @@ class SynchronousComputeService:
         else:
             self.scopes = self.settings.scopes
 
+        if self.settings.scopes_exclude is None:
+            self.scopes_exclude = []
+        else:
+            self.scopes_exclude = self.settings.scopes_exclude
+
         self.shared_basedir = Path(self.settings.shared_basedir).absolute()
         self.shared_basedir.mkdir(exist_ok=True)
         self.keep_shared = self.settings.keep_shared
@@ -168,6 +173,7 @@ class SynchronousComputeService:
 
         tasks = self.client.claim_tasks(
             scopes=self.scopes,
+            scopes_exclude=self.scopes_exclude,
             compute_service_id=self.compute_service_id,
             count=count,
             protocols=self.settings.protocols,
--- a/alchemiscale/compute/service.py
+++ b/alchemiscale/compute/service.py
@@ -97,10 +97,7 @@ class SynchronousComputeService:
         else:
             self.scopes = self.settings.scopes
 
-        if self.settings.scopes_exclude is None:
-            self.scopes_exclude = []
-        else:
-            self.scopes_exclude = self.settings.scopes_exclude
+        self.scopes_exclude = self.settings.scopes_exclude
 
         self.shared_basedir = Path(self.settings.shared_basedir).absolute()
         self.shared_basedir.mkdir(exist_ok=True)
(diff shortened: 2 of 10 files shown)


dotsdl (commented):
I'm happy with this. Ready for review @ianmkenney!

alchemiscale/compute/service.py
```diff
@@ -97,6 +97,11 @@ def __init__(self, settings: ComputeServiceSettings):
         else:
             self.scopes = self.settings.scopes
 
+        if self.settings.scopes_exclude is None:
+            self.scopes_exclude = []
+        else:
+            self.scopes_exclude = self.settings.scopes_exclude
```
ianmkenney, line 103: `client.claim_tasks` accepts `None` as a valid input. So converting it to an empty list is unnecessary. If you provide an empty list to `client.claim_tasks`, it gets converted to `None`, so you're just round-tripping a `None`. Replacing the entire block with `self.scopes_exclude = self.settings.scopes_exclude` produces passing tests. This particular `else` was not hit by the test suite.

alchemiscale/compute/settings.py
```diff
@@ -124,7 +128,24 @@ class Config:
...
+                _values[idx] = Scope.from_str(value)
+            else:
+                raise ValueError("Unable to parse input as a Scope")
+        return _values
+
+    @field_validator("scopes_exclude", mode="before")
+    @classmethod
+    def validate_scopes_exclude(cls, values) -> list[Scope] | None:
+        if values is None:
+            return None
```
ianmkenney, line 148: Repeated validation logic

alchemiscale/compute/settings.py
```diff
@@ -124,7 +128,24 @@ class Config:
 
     @field_validator("scopes", mode="before")
     @classmethod
-    def validate_scopes(cls, values) -> list[Scope]:
+    def validate_scopes(cls, values) -> list[Scope] | None:
+        if values is None:
```
ianmkenney, line 132: an empty list should be treated as `None` if we're going to bother with using `None` at all.

alchemiscale/tests/integration/compute/client/test_compute_client.py
```diff
@@ -222,6 +222,48 @@ def test_claim_tasks(
...
+        n4js_preloaded,
+        compute_client: client.AlchemiscaleComputeClient,
+        compute_service_id,
+        uvicorn_server,
+    ):
+        # register compute service id
+        compute_client.register(compute_service_id)
+
+        # claim with scopes_exclude matching scope_test — should return empty
+        # since the client token only has access to scope_test
```
ianmkenney, line 237: True, the client token only has access to scope_test, but that's not the reason it should be empty. If the client had a greater scope access this would still return an empty list.

alchemiscale/tests/integration/compute/client/test_compute_client.py
```diff
@@ -222,6 +222,48 @@ def test_claim_tasks(
...
+        task_sks = compute_client.claim_tasks(
+            scopes=[scope_test],
+            scopes_exclude=[scope_test],
+            compute_service_id=compute_service_id,
+        )
+
+        assert task_sks == []
+
+    def test_claim_tasks_scopes_exclude_none(
```
ianmkenney, line 246: This repeats `test_claim_tasks` since `None` is the default.

alchemiscale/compute/api.py
```diff
@@ -255,6 +256,15 @@ def claim_tasks(
     for single_query_scope in set(query_scopes):
         taskhubs.update(n4js.query_taskhubs(scope=single_query_scope, return_gufe=True))
 
+    # filter out taskhubs that match excluded scopes
+    if scopes_exclude:
+        scopes_exclude_reduced = minimize_scope_space(scopes_exclude)
+        taskhubs = {
```
ianmkenney, line 262: This expression is a little dense on first glance

alchemiscale/tests/integration/compute/client/test_compute_client.py
```diff
@@ -222,6 +222,48 @@ def test_claim_tasks(
             t.gufe_key for t in all_tasks if priorities[t] == 1
         ]
 
+    def test_claim_tasks_scopes_exclude(
```
ianmkenney, line 225: This only tests with a single scope and doesn't really demonstrate filtering as much as blocking.

alchemiscale/tests/integration/compute/conftest.py
ianmkenney, line 1: I think these changes are an improvement to how we build our fixtures.

alchemiscale/tests/integration/compute/test_compute_api.py
```diff
@@ -106,3 +106,102 @@ def test_get_task_transformation_bad_scope(
...
+            compute_service_id=str(compute_service_id),
+            count=1,
+            protocols=None,
+        )
+        response = multi_scope_test_client.post("/claim", json=data_with_exclude)
+        assert response.status_code == 200
+        tasks_with_exclude = response.json()
+        assert len(tasks_with_exclude) == 1
+        # the claimed task should not be from scope_test
+        if tasks_with_exclude[0] is not None:
```
ianmkenney, line 152: we already know that the value is not `None`, not sure why this is here

alchemiscale/tests/integration/compute/test_compute_api.py
```diff
@@ -106,3 +106,102 @@ def test_get_task_transformation_bad_scope(
...
+            claimed_sk = ScopedKey.from_str(tasks_with_exclude[0])
+            assert claimed_sk.scope != scope_test
```
ianmkenney, line 154: The tasks are returned from a random taskhub based on weights, which all taskhubs have the same weight, as currently implemented. Assuming the exclusion works, we know that the assertion passes. However, if exclusion doesn't work, we might randomly pass.

alchemiscale/tests/integration/compute/test_compute_api.py
```diff
@@ -106,3 +106,102 @@ def test_get_task_transformation_bad_scope(
...
+        data = dict(
+            scopes=[s.to_dict() for s in multiple_scopes],
+            scopes_exclude=[s.to_dict() for s in multiple_scopes],
+            compute_service_id=str(compute_service_id),
+            count=1,
+            protocols=None,
+        )
+        response = multi_scope_test_client.post("/claim", json=data)
+        assert response.status_code == 200
+        assert response.json() == []
```
ianmkenney, line 180: Intuitively, I'd expect this to be `[None]`, but `claim_tasks` does return an empty list. I feel like that behavior is left over from the old style of returns. This might be a bug.

https://github.com/OpenFreeEnergy/alchemiscale/commit/f78d5adc641546fb7d76145138d8789f87e2da88 introduced early return with an empty list, but https://github.com/OpenFreeEnergy/alchemiscale/commit/208eeb39481508faffde1b2332c8f7f8ba87a3d4 makes the **final** return pad the returned list with `None`.

alchemiscale/tests/integration/compute/test_compute_api.py
```diff
@@ -106,3 +106,102 @@ def test_get_task_transformation_bad_scope(
...
+        data = dict(
+            scopes=[s.to_dict() for s in multiple_scopes],
+            scopes_exclude=[wildcard_scope.to_dict()],
+            compute_service_id=str(compute_service_id),
+            count=1,
+            protocols=None,
+        )
+        response = multi_scope_test_client.post("/claim", json=data)
+        assert response.status_code == 200
+        assert response.json() == []
```
ianmkenney, line 207: see previous statement regarding expected return

alchemiscale/tests/unit/compute/test_settings.py
```diff
@@ -30,3 +30,34 @@ def test_validate_scopes(self):
         with pytest.raises(ValueError):
             scopes = ["*.*.*", "*-*-*"]
             ComputeServiceSettings.validate_scopes(scopes)
+
+    def test_validate_scopes_none(self):
+        assert ComputeServiceSettings.validate_scopes(None) is None
+
+    def test_validate_scopes_exclude(self):
```
ianmkenney, line 37: Almost an exact copy of `test_validate_scopes`

ianmkenney (changes requested):

dotsdl:
Thanks for the review @ianmkenney! Addressed in ca6dc2a:

- **`service.py` `scopes_exclude` round-trip** — dropped the `None → [] → None` block; now just `self.scopes_exclude = self.settings.scopes_exclude`.
- **Repeated validator logic** — consolidated `validate_scopes` and `validate_scopes_exclude` into a single `@field_validator("scopes", "scopes_exclude", ...)`.
- **Empty list treated as `None`** — the consolidated validator now returns `None` for both `None` and `[]`.
- **Dense filter expression in `/claim`** — pulled the predicate out into a named local `is_excluded(taskhub_scope)` helper.
- **Pad-on-empty inconsistency** — fixed the early `return []` path to `return [None] * count`, matching the normal claim-path return shape. Tests updated to assert against the padded form.
- **`test_claim_tasks_scopes_exclude` misleading comment** — rewrote: the result is empty because excluding the only included scope leaves no candidate taskhubs, not because of the client token. Asserts against `[None]` now.
- **`test_claim_tasks_scopes_exclude_none` duplicates `test_claim_tasks`** — removed.
- **Multi-scope filter test is non-deterministic** — `test_claim_tasks_with_scopes_exclude` now requests `count=100` (well above total available), exhausts every non-excluded taskhub, and asserts (a) at least one task was claimed and (b) none came from `scope_test`.
- **Redundant `if tasks_with_exclude[0] is not None`** — removed alongside the rewrite above.
- **`test_validate_scopes_exclude` duplicates `test_validate_scopes`** — collapsed into a single `test_validate_scopes_empty` parametrized over `[None, []]`.

Also added a news entry for #222.

alchemiscale/compute/api.py
dotsdl: Agreed! Simplified it!

alchemiscale/tests/integration/compute/test_compute_api.py
dotsdl: Good catch! Fixed to make this consistent with normal return behavior!

dotsdl:
Follow-up in ccdf7cb: I'd only addressed the "filtering vs blocking" point at the API layer, not at the client layer where you actually flagged it. Added a `fully_scoped_compute_client` fixture (using the existing `fully_scoped_credentialed_compute` identity) and `test_claim_tasks_scopes_exclude_filters`, which claims across all 3 scopes with one excluded, asserts at least one task is claimed, and asserts none come from the excluded scope.

dotsdl (commented):
@ianmkenney I think this is ready for another pass! Thank you for your review!

ianmkenney (approved):

PR #486, opened 2026-02-04
Add AlchemicalArchive extraction

### LINKED ISSUE #246: Add `AlchemiscaleClient` method for producing archival-quality extracts of `AlchemicalNetwork`s and their existing results
Since `alchemiscale` is an execution system for `AlchemicalNetwork`s, not an archival system, we should provide a clear pathway for producing archival-quality extracts of `AlchemicalNetwork`s and their existing results.

There are potentially many ways to accomplish this, but ideally the serialized form of an `AlchemicalNetwork` and its `ProtocolDAGResult`s has the following properties:
1. Can be easily consumed by analysis tools, such as `cinnabar`.
2. Can be used as a starting point for tools used to define `AlchemicalNetwork`s, so round-trippable into `GufeTokenizable`s with at least the same library versions as those used to create them.
3. Possible to usefully parse without any OpenFE tooling at all, even if not recommended.
4. Suitable for archival and sharing on citable [FAIR](https://en.wikipedia.org/wiki/FAIR_data) data repositories, such as Zenodo.

Some of the solution to this issue may make sense for inclusion in [`gufe`](https://github.com/OpenFreeEnergy/gufe), as this need is not necessarily specific to `alchemiscale`.

ianmkenney:
closes #246

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,85 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+**alchemiscale** is a high-throughput alchemical free energy execution system for HPC, cloud, bare metal, and Folding@Home. It's a Python 3.11+ service-oriented system using Neo4j as a graph database state store and AWS S3 for object storage.
+
+## Build & Development Commands
+
+```bash
+# Install (editable, no-deps since conda handles deps)
+pip install --no-deps -e .
+
+# Environment setup via conda/mamba
+micromamba create -f devtools/conda-envs/test.yml
+micromamba activate alchemiscale-test
+
+# Run all tests (parallel)
+pytest -n auto -v --cov=alchemiscale alchemiscale/tests
+
+# Run a single test file
+pytest alchemiscale/tests/unit/test_models.py
+
+# Run a single test by name
+pytest -k "test_name_pattern" alchemiscale/tests
+
+# Unit tests only
+pytest alchemiscale/tests/unit
+
+# Integration tests only (requires Docker for Neo4j via grolt)
+pytest alchemiscale/tests/integration
+
+# Formatting
+black alchemiscale
+black --check --diff alchemiscale   # check only (CI uses this)
+
+# Type checking
+mypy alchemiscale
+```
+
+## Architecture
+
+The system has four main services communicating through Neo4j and S3:
+
+- **AlchemiscaleAPI** (`interface/api.py`): User-facing FastAPI REST API for submitting networks, actioning tasks, retrieving results. Deployed with Gunicorn + Uvicorn.
+- **AlchemiscaleComputeAPI** (`compute/api.py`): Compute-facing FastAPI REST API for task claiming and result submission.
+- **SynchronousComputeService** (`compute/service.py`): Runs on HPC/GPU nodes, claims tasks, executes GUFE ProtocolDAGs, submits results.
+- **StrategistService** (`strategist/service.py`): Automated strategy execution service using the `stratocaster` framework.
+
+### Key Modules
+
+- `models.py`: Core data models — `Scope` (org/campaign/project hierarchy) and `ScopedKey` (globally unique identifier within a scope).
+- `storage/statestore.py`: `Neo4jStore` — the largest module, contains all Neo4j Cypher query logic for state management. This is the source of truth for the system.
+- `storage/objectstore.py`: `S3ObjectStore` — stores serialized ProtocolDAG results.
+- `base/api.py`: Shared FastAPI router logic, JWT auth, scope validation, response handling. Both APIs inherit from this.
+- `base/client.py`: Shared HTTP client with retry logic. Both `AlchemiscaleClient` and `AlchemiscaleComputeClient` inherit from this.
+- `security/auth.py`: JWT token creation/validation, password hashing with bcrypt.
+- `settings.py`: Pydantic `BaseSettings` classes that auto-populate from environment variables (Neo4j, S3, JWT, API settings).
+- `compression.py`: Zstandard compression/decompression for GUFE objects.
+- `cli.py`: Click-based CLI entry point (`alchemiscale` command) for starting APIs, compute services, database management, and identity management.
+
+### Data Flow
+
+1. Users interact via `AlchemiscaleClient` → `AlchemiscaleAPI`
+2. Tasks are created and stored in Neo4j with scope-based access control
+3. Compute services claim tasks via `AlchemiscaleComputeAPI`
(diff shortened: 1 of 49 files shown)


dotsdl:
Awaiting new `gufe` release featuring `AlchemicalArchive` before we merge.

dotsdl:
@ianmkenney would you like me to continue on this one? Otherwise happy to leave it to you!

ianmkenney:
@dotsdl you can take over if you'd like

dotsdl:
Can do! Thanks for all your work on this! Will ask you for a review once I'm finished!

dotsdl:
## Design note: client-side archive assembly

This PR builds `AlchemicalArchive` extracts **client-side** (`AlchemiscaleClient.get_network_archive` / `get_network_archives`), composing existing retrieval methods, rather than adding a server-side endpoint. The original `[DNM]` draft sketched a `/bulk/networks/archive` endpoint; we removed it and reimplemented on the client. Rationale and tradeoffs:

### Why client-side

- **API worker memory.** An archive is inherently bulk: it pulls every successful `ProtocolDAGResult` (the large objects in the system, backed by S3) for every `Transformation` in a network and serializes them into one object. Doing that server-side forces a shared API worker to hold the entire result set in memory for a single request. The workers are sized for many small, concurrent request/response cycles — archive generation is the opposite shape (few, huge, bursty). Supporting it server-side would mean provisioning every worker for a peak that is idle almost all the time, i.e. a large, near-permanent over-provision.
- **No new API-side bottleneck.** A long-lived, memory-heavy request risks head-of-line blocking for all other users of that worker. Client-side, the weight lands in the user's own process, where a big archive is their RAM, not the server's.
- **Reuse of tested machinery.** Assembly is just `AlchemicalNetwork` + per-`Transformation` `ProtocolDAGResult`s + optional metadata. The client already has cached, parallelized retrieval for exactly those pieces, so there is nothing to reimplement server-side. `AlchemicalArchive` construction is `gufe` logic that runs identically on either side, so centralizing it buys little.
- **Smaller API surface.** No new endpoint to authorize, scope-check, version, and maintain.

### Tradeoffs accepted

- **More round trips.** Client-side assembly means N per-transformation fetches instead of one request. Mitigated by using `get_network_results`, which fans the `ProtocolDAGResult` retrieval out across processes, so the dominant cost is parallel rather than serial. Acceptable for an infrequent, inherently-bulk export operation.
- **When an endpoint would earn its keep:** a concrete need the client can't meet — e.g. server-mediated archival for clients that cannot hold the full result set. No such need today; if one arises, a streaming/paginated endpoint can be added without disturbing this client API.

### Correctness note

Results are joined to `Transformation`s via their **authoritative server-side `ScopedKey`s** (from `get_network_results` / `get_network_transformations`), then each `Transformation` is fetched by that `ScopedKey`. We deliberately do **not** reconstruct a `ScopedKey` from a freshly-deserialized `Transformation`'s `GufeKey`: a `ScopedKey` carries the ingestion-time `gufe_key`, but a `GufeTokenizable` deserialized later may recompute a different `GufeKey` if `gufe` tokenization has changed since ingestion — so a key-reconstructed join can silently drop results rather than error.

dotsdl (approved):
I'm happy with this! Thanks for the initial work on this @ianmkenney!

alchemiscale/interface/client.py
```diff
@@ -1767,6 +1765,174 @@ def get_network_failures(
...
+        )
+
+        # pair each Transformation with its results by fetching the
+        # Transformation via its authoritative ScopedKey. We deliberately do
+        # not derive a ScopedKey from a deserialized Transformation's GufeKey:
+        # a `Transformation` deserialized now may not reproduce the GufeKey it
+        # had when ingested (gufe tokenization can change across versions), so
+        # its key is not a reliable join to the stored ScopedKey.
+        transformation_results = []
+        for transformation_sk, pdrs in results.items():
```
mikemhenry, line 1939: I think we should make the `transformation_results` ordering deterministic before constructing the `AlchemicalArchive.get_network_results()` builds its result dict from `as_completed(...)`, so `results.items()` here reflects process completion order rather than network order or key order. That means two archive exports for the same network/results could produce different serialized archives. Could we iterate in a stable order, e.g. `for transformation_sk in sorted(results, key=str): ...,` or preserve the order from `get_network_transformations(network)`? 

I think it makes sense to leave `get_network_results(...)` as is (since I think in general we should not keep order preserved since there could be performance reasons to keep it unsorted) and instead sort when we iterate over the results.

CLAUDE.md
```diff
@@ -75,6 +75,7 @@ The system has four main services communicating through Neo4j and S3:
...
 - **GUFE integration**: Core chemistry types (`AlchemicalNetwork`, `Transformation`, `ChemicalSystem`, `Protocol`) come from the `gufe` library and are stored/retrieved via their tokenization system (`GufeKey`, `GufeTokenizable`).
+- **`GufeKey` is not stable across `gufe` versions**: A `ScopedKey` is stable once created (it carries the *ingestion-time* `gufe_key`), but a `GufeTokenizable` deserialized later may recompute a *different* `GufeKey` if the `gufe` tokenization has changed since ingestion. Never derive a `ScopedKey` from a freshly-deserialized object's `.key` (e.g. `ScopedKey(gufe_key=obj.key, **scope.to_dict())`) to join against stored data — such a join can silently miss instead of erroring. Instead, obtain the authoritative `ScopedKey` from the server (e.g. `get_network_transformations`, or the keys of `get_network_results`) and fetch the object *by that `ScopedKey`* (e.g. `get_transformation(sk)`). Comparisons purely among objects deserialized in the same process (e.g. "is this transformation in `network.edges`?") are safe, since both sides use the current-stack key consistently.
```
mikemhenry, line 78: Does this just keep a running list of tips for Claude?

mikemhenry (changes requested):

CLAUDE.md
mikemhenry: This was supposed to be a question, this isn't a blocking comment.

alchemiscale/interface/client.py
dotsdl: `AlchemicalArchive` already sorts `transformation_results` appropriately for exactly this reason, so we should be good there: https://github.com/OpenFreeEnergy/gufe/blob/8a928536613f13b8b7fe7c357aa53c0dec13bdd6/src/gufe/archival.py#L70

CLAUDE.md
dotsdl: Yes! As I worked with Claude on this I noted this subtlety in how we use `ScopedKey`s, and asked it to add this to `CLAUDE.md` so it's built-in knowledge when working with this repo.

PR #487, opened 2026-02-26
Fix async event loop incompatibility with async-lru 2.2.0

dotsdl:
## Summary
- `async-lru` 2.2.0 (released Feb 20) added strict event loop checks that broke `@alru_cache` usage — `asyncio.run()` creates a new loop each time, but `alru_cache` now rejects cross-loop calls
- Replaced all three `asyncio.run()` call sites with `_run_async()`, a new method on `AlchemiscaleBaseClient` that maintains a single shared event loop per class, matching `alru_cache`'s class-level loop tracking
- `nest_asyncio` applied proactively for Jupyter compatibility

## Test plan
- [x] All 138 unit tests pass
- [x] All 376 integration tests pass (including previously-failing `test_cached_pdr` and `test_get_transformation_and_network_results`)

### DIFF
--- a/alchemiscale/base/client.py
+++ b/alchemiscale/base/client.py
@@ -116,6 +116,7 @@ class AlchemiscaleBaseClient:
 
     _exception = AlchemiscaleBaseClientError
     _retry_status_codes = [404, 502, 503, 504]
+    _shared_event_loop = None
 
     _PARAMS = {
         "api_url": AlchemiscaleBaseClientParam(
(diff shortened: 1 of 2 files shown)


dotsdl:
@ianmkenney can you review this?

alchemiscale/base/client.py
```diff
@@ -262,6 +263,28 @@ def _settings(self):
...
+        ``alru_cache`` binds to the event loop it is first used with and rejects
+        calls from different loops. ``asyncio.run()`` creates a new loop each time,
+        which is incompatible. This method maintains a single shared event loop
+        per class to match ``alru_cache``'s class-level loop tracking.
+
+        We apply ``nest_asyncio`` proactively so the loop also works in
+        environments where an event loop is already running (e.g. Jupyter).
+
+        """
+        import nest_asyncio
```
ianmkenney, line 278: Since this is now being applied proactively, I'll just note that it looks like this library is no longer maintained.

ianmkenney (approved):
Just a note on nest_asyncio

alchemiscale/base/client.py
dotsdl: I'd like to replace `nest_asyncio` if possible in a future PR. I'll make an issue for this!

PR #502, opened 2026-05-04
Move autoscaling sizing logic into ComputeManager base

dotsdl:
## Summary

Centralize the per-cycle "how many compute services should we create?" calculation in `ComputeManager`. Subclasses today duplicate the same arithmetic — `alchemiscale-k8s`'s `K8SManager` does it inline, and `alchemiscale-hpc`'s `ScriptTemplateHPCManager` recently ported the same formula. Any future backend would need to do the same.

The four inputs are all general autoscaling concepts; none of them are backend-specific:

- `num_tasks` (server-reported on the OK instruction)
- `max_compute_services` (already on `ComputeManagerSettings`)
- `claim_limit` (already on `ComputeServiceSettings`)
- per-cycle submission cap (was K8s `job_creation_rate` / HPC `max_submit_per_cycle`; upstreamed here as `max_submit_per_cycle` on `ComputeManagerSettings`)

So this PR adds `_compute_jobs_to_create(num_tasks, num_active_services) -> int` as a concrete method on `ComputeManager`, has `cycle()` invoke it, and passes the resulting `target` to `create_compute_services`. The signature of the abstract method becomes `create_compute_services(data, target) -> int`.

## Why pass `target` rather than just exposing the helper

Two options on the table:

1. Add `_compute_jobs_to_create` as a helper, leave `create_compute_services(data)` alone. Subclasses *can* call it; nothing forces them to.
2. Compute it in `cycle()` and pass it as `target`. Subclasses *must* honor (or ignore) the target, but the math is provably called once per cycle.

Option 2 is stronger:
- Removes the burden of remembering to call the helper
- Keeps the math in one place rather than scattered across N subclasses
- Makes the subclass interface narrower: "submit up to N jobs after doing your backend-specific gating"
- Logs the sizing decision once, in the base, not duplicated across backends

Option 1 is barely better than the status quo.

## Default sizing rule

```
min(num_tasks, max_submit_per_cycle, remaining_capacity) // claim_limit
```

with a floor of 1 when the cycle has decided we should be scaling up (i.e. `num_tasks > 0` and `remaining_capacity > 0`). The floor handles the case where the `claim_limit` divide collapses to zero (e.g. one task with `claim_limit=2`): rather than stalling, create one service and let the next cycle catch up.

Subclasses should rarely override `_compute_jobs_to_create`. Backend-specific gating (e.g. "skip this cycle if any submitted jobs are still pending in the batch system") belongs in `create_compute_services`, not in the sizing math.

`cycle()` now logs the decision each cycle:

```
Sizing: num_tasks=20, max_submit_per_cycle=5, remaining_capacity=10, claim_limit=4 -> target=5
```

## Breaking change

`ComputeManager.create_compute_services` signature: `(self, data: dict) -> int` → `(self, data: dict, target: int) -> int`.

In-tree, `LocalTestingComputeManager` (the integration test fixture) is migrated to honor `target` by spawning `range(target)` services rather than always one. Existing test expectations still hold under the new sizing because with `max_submit_per_cycle=1` and `claim_limit=2`, `target=1` each cycle for the 3-task fixture.

Out-of-tree, follow-up PRs against `alchemiscale-k8s` and `alchemiscale-hpc` will drop their local sizing math and update their `create_compute_services` signatures. Both are small mechanical changes.

## Tests

New unit tests in `tests/unit/compute/test_manager.py` exercise:

- Each cap individually (`num_tasks`, `max_submit_per_cycle`, `remaining_capacity`)
- The `claim_limit` divisor
- The floor-to-1 rule
- The zero-`claim_limit` guard (treated as 1; no `ZeroDivisionError`)
- The at-capacity short-circuit (returns 0; floor does NOT kick in)
- Combined caps via `min()`
- The new `max_submit_per_cycle` setting default and override

13/13 unit tests pass without Neo4j or S3. Full unit suite (150 tests) green.

## Test plan

- [x] `pytest alchemiscale/tests/unit/compute/` — 13 new tests pass
- [x] `pytest alchemiscale/tests/unit/` — 150 tests pass (existing 137 + new 13)
- [x] `pytest alchemiscale/tests/integration/compute/client/test_compute_manager.py` — migrated `LocalTestingComputeManager` still passes the existing scenarios
- [x] CI passes
- [ ] Coordinate follow-up PRs against `alchemiscale-k8s` and `alchemiscale-hpc` once this lands

### DIFF
--- a/alchemiscale/compute/manager.py
+++ b/alchemiscale/compute/manager.py
@@ -226,7 +226,7 @@ class ComputeManager:
                         f"-> target={target}"
                     )
                     if target > 0:
-                        new_services = self.create_compute_services(data, target=target)
+                        new_services = self.create_compute_services(data, target)
                         total_services += new_services
                         if new_services:
                             self.logger.info(
--- a/alchemiscale/compute/manager.py
+++ b/alchemiscale/compute/manager.py
@@ -105,12 +105,88 @@ class ComputeManager:
             self.logger.info(f"Deregistration successful")
 
     @abstractmethod
-    def create_compute_services(self, data: dict) -> int:
-        """Method responsible for creating compute services based on
-        data returned with an OK ComputeManagerInstruction. This must
-        return the number of compute services started.
+    def create_compute_services(self, data: dict, target: int) -> int:
+        """Submit up to ``target`` new compute services and return the actual
+        number created.
+
+        Called by :meth:`cycle` when scaling up is warranted. ``target`` is the
+        per-cycle sizing decision computed by :meth:`_compute_jobs_to_create`,
+        which already accounts for the number of waiting tasks, the per-cycle
+        rate limit (``max_submit_per_cycle``), the remaining capacity
+        (``max_compute_services - len(compute_service_ids)``), and the
+        per-service ``claim_limit``.
+
+        Subclasses still own backend-specific gating (e.g. checking whether
+        any previously submitted jobs are still pending in the batch system,
+        running health checks). ``target`` is an upper bound, not a guarantee.
+
+        Parameters
+        ----------
+        data
+            Payload from the OK ``ComputeManagerInstruction``. Includes
+            ``compute_service_ids`` and ``num_tasks`` at minimum.
+        target
+            Upper bound on the number of services to create this cycle.
+            Always ``>= 1`` when this method is called; the cycle short-
+            circuits and never calls in if there is no work or no capacity.
+
+        Returns
+        -------
+        int
+            Number of compute services actually created.
         """
         raise NotImplementedError
+
+    def _compute_jobs_to_create(self, num_tasks: int, num_active_services: int) -> int:
+        """Decide how many new compute services to create this cycle.
+
+        The formula:
+
+            ``min(num_tasks, max_submit_per_cycle, remaining_capacity)
+            // claim_limit``
+
+        with a floor of one when the cycle has decided we should be scaling
+        up (i.e. ``num_tasks > 0`` and ``remaining_capacity > 0``). The
+        floor handles the case where the ``claim_limit`` divide collapses
+        to zero (e.g. a single task with ``claim_limit=2``): rather than
+        stalling, create one service and let the next cycle catch up.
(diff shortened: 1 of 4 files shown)


dotsdl (commented):
I'm happy with the content on this one. @ianmkenney this upstreams the sizing logic we implemented in `alchemiscale-k8s` for determining how many jobs to submit given multiple factors (available tasks, running services, claim limit). I will propagate this change to `alchemiscale-k8s` and `alchemiscale-hpc` after merge.

Can I get your review on this?

alchemiscale/compute/manager.py
```diff
@@ -137,14 +213,31 @@ def cycle(self) -> bool:
...
+                    self.logger.info(
+                        f"Sizing: num_tasks={num_tasks}, "
+                        f"max_submit_per_cycle={self.settings.max_submit_per_cycle}, "
+                        f"remaining_capacity="
+                        f"{self.settings.max_compute_services - total_services}, "
+                        f"claim_limit={self.service_settings.claim_limit} "
+                        f"-> target={target}"
+                    )
+                    if target > 0:
+                        new_services = self.create_compute_services(data, target=target)
```
ianmkenney: keyword arg syntax feels a bit unnecessary here, especially when it's just one arg

alchemiscale/tests/unit/compute/test_manager.py
```diff
@@ -0,0 +1,129 @@
...
+
+Focuses on the pure-arithmetic sizing logic in ``_compute_jobs_to_create``,
+which can be exercised without standing up Neo4j, S3, or a running compute
+API. Integration coverage of the cycle loop lives in
+``tests/integration/compute/client/test_compute_manager.py``.
+"""
+
+from unittest.mock import MagicMock
+
+import pytest
```
ianmkenney: unused

ianmkenney (approved):
The logic looks sound and the change overall change makes sense.

dotsdl:
Awesome, thanks so much @ianmkenney! Merging!

PR #503, opened 2026-05-28
Improve robustness of service termination

dotsdl:
## Summary

Makes the long-running services — `SynchronousComputeService`, `ComputeManager`, and `StrategistService` — shut down **promptly and cleanly** when asked to stop, instead of ignoring `stop()` (and termination signals) until the current sleep interval elapses. Along the way it fixes a latent dead mechanism, removes dead code, and consolidates the sleep primitive into one shared home.

## Motivation

All three services run a `while`-loop that sleeps between work cycles. They relied on a plain, **uninterruptible** `time.sleep()`, so a `stop()` (e.g. from a SIGTERM/SIGINT handler) was not noticed until the sleep finished on its own — up to `deep_sleep_interval` (300s) for the compute service, `sleep_interval` (default 1800s) for the manager. For services that hold a registration in the state store, a slow or skipped shutdown means an **orphaned `ComputeManagerRegistration`/`ComputeServiceRegistration`**.

## Changes

**Interruptible sleep across all three services**
- `ComputeManager` now sleeps via the `InterruptableSleep` functor, woken by `stop()`.
- `SynchronousComputeService` *constructed* an `InterruptableSleep` and interrupted it in `stop()`, but every actual sleep used plain `time.sleep()` — the functor was never called and the `except SleepInterrupted` handler was dead code. Its cycle sleeps now go through it.
- `StrategistService` sleeps between cycles via the functor too. **No `KeyboardInterrupt` is raised** for it: its cooperative `_stop` checks and `ProcessPoolExecutor` teardown are preserved, since raising asynchronously into the pool-managing thread risks orphaning child processes.
- Each `start()` clears the sleep event so a stopped service can be cleanly restarted.

**Shared primitive**
- Moved `InterruptableSleep`/`SleepInterrupted` into a new dependency-free `alchemiscale.sleep` module (a third consumer in a different subpackage made the previous home in `compute.service` an awkward coupling). Re-exported from `alchemiscale.compute.service` for backwards compatibility.

**Signal-handling helper**
- Added `alchemiscale.compute.signals.install_stop_handlers`, a small reusable helper that wires `SIGHUP`/`SIGINT`/`SIGTERM` to a service's `stop()`. Signal disposition is process-global and main-thread-only, so it stays at the entry point — the helper just removes the per-CLI boilerplate (and the risk of forgetting it). The compute service `synchronous` CLI command now uses it.

**Cleanup**
- Removed the unused `sched.scheduler` instances from `SynchronousComputeService`/`AsynchronousComputeService` and the now-unused `import sched`; corrected the `InterruptableSleep` docstring, which described a scheduler integration that was never wired up.

**Tests**
- Added tests asserting `stop()` promptly wakes a sleeping manager, compute service, and strategist, plus unit tests for `install_stop_handlers`.

## Companion PRs (separate repos)

The autoscaling **manager CLIs** in `alchemiscale-hpc` (SLURM) and `alchemiscale-k8s` install no signal handlers today, so SIGTERM — the signal `systemd`/`docker stop`/`scancel`/Kubernetes actually send — bypasses the `finally` block and orphans the manager registration. Those CLIs will be wired to `install_stop_handlers` in follow-up PRs to their respective repos (they consume the helper added here).

### DIFF
--- a/alchemiscale/compute/service.py
+++ b/alchemiscale/compute/service.py
@@ -20,43 +20,8 @@ from .settings import ComputeServiceSettings
 from ..storage.models import ComputeServiceID
 from ..models import Scope, ScopedKey
 
-
-class SleepInterrupted(BaseException):
-    """
-    Exception class used to signal that an InterruptableSleep was interrupted
-
-    This (like KeyboardInterrupt) derives from BaseException to prevent
-    it from being handled with "except Exception".
-    """
-
-    pass
-
-
-class InterruptableSleep:
-    """
-    A class for sleeping, but interruptable
-
-    This class uses threading Events to wake up from a sleep before the entire sleep
-    duration has run. If the sleep is interrupted, then an SleepInterrupted exception is raised.
-
-    This class is a functor: call an instance with a delay in seconds to sleep for that
-    duration (e.g. ``int_sleep(30)``), and call :meth:`interrupt` from another thread to
-    wake it early.
-    """
-
-    def __init__(self):
-        self._event = threading.Event()
-
-    def __call__(self, delay: float):
-        interrupted = self._event.wait(delay)
-        if interrupted:
-            raise SleepInterrupted()
-
-    def interrupt(self):
-        self._event.set()
-
-    def clear(self):
-        self._event.clear()
+# moved to alchemiscale.sleep; re-exported here for backwards compatibility
+from ..sleep import InterruptableSleep, SleepInterrupted
 
 
 class SynchronousComputeService:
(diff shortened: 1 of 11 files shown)


dotsdl (commented):
I'm happy with this! Ready for review @ianmkenney, @atravitz, or @mikemhenry!

alchemiscale/compute/service.py
mikemhenry, line 409: If there is an exception before this point, `self._register()` will not get the deregistered with `self._deregister()`. I think for this class and `ComputeManager.start()` you should do all the registering also within the `try` block so we can be sure nothing gets left behind

mikemhenry (dismissed):
Just a few suggestions!

Also, I think a test (even a smoke test) for the CLI would help with coverage and also make sure the changes to the CLI are doing what we expect them to.

mikemhenry:
Okay pushed two tests and a quick fix that should work, my plan is to make a context manager that should make this easier

dotsdl (commented):
I'm happy with this, thank you @mikemhenry! Merging once CI passes one last time!

PR #507, opened 2026-06-08
Add copy_network, merge_networks, merge_scopes (closes #221)

### LINKED ISSUE #221: Add `AlchemiscaleClient.copy_network`, `.merge_networks`, and `.merge_scopes` methods

dotsdl:
## Summary

Revives the work originally proposed in #433 (now unreachable since the contributing fork was deleted) and finishes it.

- Adds ``Neo4jStore.merge_networks`` for combining multiple existing ``AlchemicalNetwork``\s into a new ``AlchemicalNetwork`` in a target ``Scope``, with completed and errored ``Task``\s (and their ``ProtocolDAGResultRef``\s) cloned into the new scope so previously-computed results do not need to be re-run. *(Originally authored by @ianmkenney; preserved here as-is, then refactored — see commits.)*
- Adds the user-facing surface:
  - ``POST /networks/merge`` on ``AlchemiscaleAPI``, which validates the destination ``Scope`` and each source network's ``Scope`` against the caller's token before delegating to the store.
  - ``AlchemiscaleClient.merge_networks(networks, name, scope)``, with client-side guards for wildcard scopes, empty input, and non-``AlchemicalNetwork`` ``ScopedKey``\s.
- Refactors the database-layer implementation to use ``KeyedChain.decode_subchains`` (gufe ≥1.8.0), which replaces a hand-rolled ``kc_to_gufe`` helper plus manual chain-slicing loop, and lifts the inline ``TransformationData`` dataclass to a private module-level ``_TransformationData`` for readability.
- Bumps the ``gufe`` pin to ``>=1.8.0`` in ``devtools/conda-envs/test.yml`` and ``devtools/conda-envs/alchemiscale-server.yml`` (the only env files that actually exercise this code path).
- Adds integration tests at both the API and client layers covering: happy-path edge union, scope-based authorization (bad destination *and* bad source), and full preservation of ``complete``/``error`` ``Task``\s with their ``ProtocolDAGResultRef``\s — including a reachability check via ``get_network_tasks(merged_sk)``.

Closes #221.

## Notes for reviewers

- The store-level ``merge_networks`` body is unchanged in behavior from the original PR; the diff against #433's last revision is the ``decode_subchains`` refactor plus the lift of ``TransformationData``.
- The ``get_network_tasks`` assertion in ``test_merge_networks_preserves_tasks_and_results`` exercises the standard ``(:AlchemicalNetwork)-[:DEPENDS_ON]->(:Transformation)<-[:PERFORMS]-(:Task)`` traversal on the merged network. If it fails, that indicates a missing ``PERFORMS`` edge in ``_TransformationData.to_subgraph`` — flagged here for follow-up rather than fixed in this PR, since the original implementation has the same behavior.
- ``subchain_cache`` is preserved with cleaner contents (``KeyedChain.from_gufe(transformation)``) but its necessity is an open question tied to the ``PERFORMS`` follow-up above.

## Test plan

- [x] ``pytest -n auto -v alchemiscale/tests/integration/storage/test_statestore.py -k merge_networks``
- [x] ``pytest -n auto -v alchemiscale/tests/integration/interface/test_api.py -k merge_networks``
- [x] ``pytest -n auto -v alchemiscale/tests/integration/interface/client/test_client.py -k merge_networks``
- [x] ``black --check --diff alchemiscale``
- [x] ``mypy alchemiscale``
- [x] Confirm whether ``test_merge_networks_preserves_tasks_and_results`` reveals the suspected missing ``PERFORMS`` wiring; open a follow-up if so.


dotsdl:
Thanks for the careful review. Pushed `d0bf4ec` addressing the feedback.

**Blocking**

- [x] **Missing `PERFORMS` edge** — fixed in `_TransformationData.to_subgraph`. The cloned `Task` is now wired to `tf_node` with a `PERFORMS` relationship in the same iteration that creates it. This also retires the open question about `subchain_cache`: `tf_node` is now meaningful as the `PERFORMS` anchor, so the cache stays.
- [x] **Cloned tasks not actioned to the new TaskHub** — documented as intentional in both the store-level docstring and the client docstring, with the remediation path spelled out (call `action_tasks` against the merged network's TaskHub after the merge if you want errored tasks back in the queue). Holding behavior change for a follow-up.

**Should-fix**

- [x] **O(N²) dedup** — switched to `dict[GufeKey, _TransformationData]` keyed on `transformation.key`. Insertion order is preserved by the dict itself; `transformation_data.values()` replaces the prior list in downstream iteration.
- [x] **`state` parameter** — threaded through `Neo4jStore.merge_networks` → `POST /networks/merge` → `AlchemiscaleClient.merge_networks`, defaulting to `NetworkStateEnum.active`. Added a parametrized client test `test_merge_networks_respects_state` over `{active, inactive}`.

**Nits picked up**

- [x] `/networks/merge` is now `async def` for consistency with `create_network`.
- [x] Improved the `KeyError` handler to collect all missing source `ScopedKey`s before raising, instead of reporting only the first failure.
- [x] Added a one-line comment at the `zip(database_keys, transformations)` site pointing at `decode_subchains`'s yield-order guarantee.
- [x] Dropped the unused `network_tyk2` fixture from `test_merge_networks_bad_scope` and `test_merge_networks_bad_source_scope`.

**Deferred**

- `update_task_trees` staticmethod → module-level helper: skipping for this PR to keep the diff focused. Easy follow-up.
- Stray `f"..."` strings without placeholders in `test_statestore.py`: these are in the original contributor's code; cleaning them up here would muddy the diff. Happy to do it in a follow-up cosmetic PR.

**Sanity check the reviewer asked for**

Static analysis confirms the PERFORMS edge is now emitted inside the per-task loop with the correct direction (`Task` → `Transformation`) and scope properties. Will report back once I've run `pytest -n auto alchemiscale/tests/integration -k merge_networks` in an env with `grolt`.

dotsdl (commented):
I'm satisfied with this! This is looking really awesome, and I'm excited to finally merge this feature!

dotsdl:
Would you be willing to give a review on this @ianmkenney?

ianmkenney:
Unfortunately I don't think I have the bandwidth to review this PR right now.

dotsdl:
All good, no worries @ianmkenney! @mikemhenry are you willing to give this a look?

mikemhenry:
Few notes (used GPT-5.5 High to review PR):

Could we also enforce `ScopedKey.qualname == "AlchemicalNetwork"` at the API and/or store layer? The client guards this, but direct REST callers can still pass a Transformation/NonTransformation scoped key. For `merge_networks` that appears capable of producing a merged network from a non-network root; for `copy_network` it could either create malformed graph state or fail with a 500-ish error path. A direct API test for `/networks/merge` with a Transformation SK and `/networks/{tf_sk}/copy` would make this explicit.

The complete-only policy may leak non-complete EXTENDS ancestors. `update_task_trees(..., task_statuses=["complete"])` filters the matched `task`, but `OPTIONAL MATCH (task)-[:EXTENDS]->(extended_task:Task)` does not filter `extended_task`. Then `to_subgraph()` clones `extended_task` and emits EXTENDS. A complete retry task extending an errored task would therefore copy the errored parent into the target scope, and that parent would not get PERFORMS unless it was also selected as a carried task. Could we either filter/gate `extended_task` to carried statuses, or explicitly document and test that ancestry tasks are copied too and ensure they get PERFORMS?

Should `merge_scopes()` preserve source network state? It queries networks with `state=None`, but each `copy_network()` call uses the default `state=active`. For scope migration/admin workflows, silently turning inactive/deleted networks active in the target scope seems surprising. Maybe fetch `get_network_state(network_sks)` and pass the corresponding state into each copy, or add a `state=` / `preserve_state=True` option and document the default.

dotsdl:
Thanks for the careful review @mikemhenry. All three points addressed across `a31c8e5` and `7ccd708`.

**1. Store/API-level qualname guard** — `a31c8e5`

Pushed the check into `Neo4jStore.merge_networks` and `Neo4jStore.copy_network` so it protects both HTTP and direct Python callers, not just the client. Rejects any non-`AlchemicalNetwork` `ScopedKey` with `ValueError`, which the API layer translates to 422. Direct API tests added: `test_merge_networks_bad_qualname`, `test_copy_network_bad_qualname`.

**2. EXTENDS ancestor leak** — `7ccd708`

Confirmed and fixed. `update_task_trees` now attaches `WHERE extended_task.status IN $task_statuses` to the `OPTIONAL MATCH`, so for a non-carried parent, `extended_task` is null-ed out without dropping the primary Task's row. `to_subgraph` then skips the parent clone and the EXTENDS edge. Went with the "gate + drop EXTENDS" option (a) rather than option (b) — cleanest match for the "only complete work is preserved" policy, and keeps the target-scope graph free of orphan `Task` nodes.

Test `test_copy_network_drops_non_complete_extends_ancestor` seeds an `error` parent + `complete` retry, copies to a target scope, and asserts (i) the errored parent isn't cloned, (ii) the retry is wired via `PERFORMS`, (iii) no dangling `EXTENDS` remains, (iv) only the retry's ok `PDRR` lands in the target scope. Doc page and news entry updated to spell out the refined `EXTENDS` semantics.

**3. `merge_scopes` state preservation** — `a31c8e5`

`AlchemiscaleClient.merge_scopes` now calls `get_networks_state` per source scope and threads each source's state through to `copy_network`. Networks in `inactive`, `invalid`, or `deleted` state on a source scope land in the target scope with the matching state, not silently reactivated. Test `test_merge_scopes_preserves_source_state` seeds one source-scope network per state and asserts each round-trips.

Non-atomicity documented in the docstring and user-guide page: fail-fast with a note that rerunning after partial failure is safe because `copy_network` is idempotent.

CI: the failing Python 3.12 job was an unrelated `test_compute_api` / `test_compute_synchronous` subprocess-boot flake (`RuntimeError: The test server could not be reached.`); 3.11 and 3.13 passed on the same run. Fresh CI kicking off on `7ccd708`.

dotsdl:
Correction on point #2 (thanks to @dotsdl for catching this): the `error → complete` shape my `7ccd708` guarded against is unreachable in practice, so I've reverted it in `ca37a68`.

Reasoning: a child `Task` C extending parent P can only reach `complete` via `waiting → running → complete`, and the `waiting → running` transition goes through the claim path — which gates on `WHERE other_task.status = 'complete' OR other_task IS NULL` (`statestore.py:3230`). So C being `complete` implies P was `complete` at claim time. From `complete`, P's only escape is `invalid` / `deleted`, and both cascade down `EXTENDS` descendants (`statestore.py:4276`, `statestore.py:4330`), taking C out of `complete` along with P.

Net: under system-enforced invariants, no `complete` Task can have a non-`complete` `EXTENDS` parent — so `update_task_trees` cannot see the leaky shape. The guard I added was solving a non-problem, and its test had to poke around the API to force an impossible state.

`a31c8e5` still stands. So all three of your review items land as: #1 fixed, #2 diagnosed as unreachable and not fixed, #3 fixed.

mikemhenry:
Some more feedback (used GPT-5.6 High to review PR):

The `AlchemicalNetwork` qualname validation now appears to be handled appropriately at the store/API level, and the changes to preserve network state during `merge_scopes()` address the main issue raised there.

I still have a few concerns:

## 1. Copied tasks may retain source-scope `extends` values

When a task node is copied, its scope properties and `_scoped_key` are rewritten for the target scope. However, it looks like the serialized `Task.extends` property is copied unchanged, meaning it may still contain the parent task’s source-scope `ScopedKey`.

This is potentially observable through `get_transformation_tasks(..., return_as="graph")`, which appears to construct the task graph using the serialized `extends` property rather than the Neo4j `EXTENDS` relationship.

This could result in an inconsistent graph where:

* the copied child task is in the target scope;
* the copied parent task is in the target scope;
* the Neo4j `EXTENDS` relationship correctly connects the copied tasks;
* but the returned task graph points from the target-scope child to the source-scope parent.

Could we either:

1. rewrite the serialized `extends` value when copying the task node; or
2. construct the returned task graph from the Neo4j `EXTENDS` relationship?

It would also be useful to extend the copy test to call `get_transformation_tasks(..., return_as="graph")` on the copied transformation and verify that all nodes and edges reference only the target scope. The current relationship-direction test does not appear to exercise this public behavior.

## 2. The invariant used to justify reverting the `EXTENDS` status check is not enforced by the store API

I understand the reasoning that a claimed child task should only run after its parent is complete, and that invalidating or deleting a parent propagates to its descendants. That appears valid for the normal compute-service claim path.

However, I do not think this is currently a store-level invariant. `set_task_running()` can transition a waiting task to running without checking whether its `EXTENDS` parent is complete, and `set_task_complete()` can subsequently make that child complete. `_validate_extends_tasks()` also appears to reject only invalid or deleted parents, not waiting, running, or errored parents.

That means a complete child extending a non-complete parent can be created through the public store methods without directly modifying Neo4j. The copy query would select the complete child and then copy its unfiltered parent as an ancestry node.

I think this should be resolved in one of the following ways:

* enforce the parent-complete requirement in `set_task_running()`;
* detect inconsistent task ancestry during copy/merge and raise an error; or
* explicitly support copying ancestry tasks regardless of their status, and document and test that behavior.

As written, the statement that this state is unreachable through system-enforced invariants seems stronger than what the store currently guarantees.

## 3. Conflicting network states during `merge_scopes()` need a defined policy

The updated implementation now preserves the source network state when copying, which addresses the original concern for individual networks.

There is still an ambiguous case when the same `AlchemicalNetwork` exists in multiple source scopes with different states. Those networks deduplicate to the same target-scope network key, so both source states cannot be preserved. It looks like whichever source is processed last would overwrite the target `NetworkMark` state.

Could this behavior be explicitly defined and tested? Possible policies include:

* reject conflicting source states;
* use a documented state precedence;
* preserve an already-existing target state; or
* guarantee and document a deterministic last-source-wins rule.

Also, the response mentions testing one network per state, but the added test appears to cover `active`, `inactive`, and `deleted`, rather than all four states including `invalid`. The implementation may already handle `invalid` generically, but the description of the test coverage should be adjusted or the missing state added.

## 4. Empty network lists are still accepted below the client layer

The client rejects `merge_networks([])`, but the REST endpoint and `Neo4jStore.merge_networks()` do not appear to enforce the same requirement.

A direct API or store caller could therefore request a merge with no source networks and potentially create an empty `AlchemicalNetwork`.

Could the non-empty validation be moved or duplicated in the store layer, with the client validation retained for quicker feedback?

The first two items seem important to resolve before merging, particularly the stale serialized `extends` value, since that could cause the stored graph and the graph returned through the public API to disagree.

dotsdl:
Thanks for the second pass @mikemhenry. Addressed all four in `135fd62`.

**1. Stale `Task.extends` on copied nodes** — confirmed and fixed. `to_subgraph` now rewrites each cloned Task's serialized `extends` property: to the target-scope parent's ScopedKey when the parent was carried into the target, or to `None` when it was not. `get_transformation_tasks(..., return_as="graph")` on a copied Transformation now returns a mapping wholly in the target scope. Extended `test_copy_network_partial_preload_target` to round-trip through `get_transformation_tasks` and check every key/value.

**2. `EXTENDS`-status invariant argument** — you're right, the "no complete child of a non-complete parent" invariant is only enforced by `claim_taskhub_tasks`'s parent-status gate, not by `set_task_running` itself, so a direct store-API caller can force the shape. Restored the `extended_status_clause` on the `OPTIONAL MATCH` in `update_task_trees`. For any non-carried parent, `extended_task` is null-ed without dropping the primary Task's row, so `to_subgraph` skips the parent clone and the `EXTENDS` edge. Combined with #1's rewrite, no non-carried Task or dangling reference can leak regardless of source-graph shape. New test `test_update_task_trees_invariant_drops_non_complete_extends_parent` forces the adversarial shape via a direct `set_task_running` bypass and verifies no leak.

**3. Conflicting network states in `merge_scopes`** — @dotsdl chose to simplify: the merge/copy machinery now works over `active` networks only. `Neo4jStore.merge_networks` and `Neo4jStore.copy_network` reject a non-`active` source with `ValueError` (surfacing as 422 through REST). `AlchemiscaleClient.merge_scopes` filters at query time so non-active source networks are skipped without error. Consolidating a scope shouldn't silently reactivate a soft-deleted or invalidated network, and there's no useful work to do on one anyway. Side-steps the cross-scope state-conflict question entirely: `merge_scopes` never sees a non-active network. `test_merge_scopes_only_copies_active_networks` covers all four `NetworkStateEnum` values on a source scope and asserts only the `active` one lands in the target.

**4. Empty `network_scoped_keys` at the store** — added `if not network_scoped_keys: raise ValueError(...)` in `Neo4jStore.merge_networks`, mirroring the client's existing check. `test_merge_networks_empty_list` at the API level covers the 422 path; `test_merge_networks_rejects_empty_list` at the store level covers the direct-Python path.

Diff: 7 files, +426 / -67 (much of it new tests and the rewrite-extends helper). Passing black; CI running.

mikemhenry:
Some more feedback (used GPT-5.6 High to review PR):

The previously raised issues around:

* stale serialized `Task.extends` values,
* copying non-complete `EXTENDS` ancestors,
* conflicting source network states,
* empty merge lists, and
* non-network scoped keys

now appear to be substantively addressed in the code, with targeted tests covering the intended behavior. CI is also passing.

I still recommend one additional change before merging.

## Preserve the state of an existing target network

`merge_scopes()` now considers only active source networks, which resolves the earlier source-state conflict. However, an active source network can still deduplicate onto a matching network that already exists in the target scope with a non-active state.

For example:

1. A network is active in the source scope.
2. The same network already exists in the target scope.
3. The target network is marked `inactive`, `invalid`, or `deleted`.
4. `merge_scopes()` is run again.

`merge_scopes()` calls `copy_network()` with the default `state=active`. Because the target `NetworkMark` already exists and is merged by identity, its state appears to be overwritten with `active`.

The current tests do not exercise this case:

* the active-only test checks non-active source networks;
* the deduplication test uses a target network that is already active.

Could we add a regression test where the matching target network is first marked non-active, then `merge_scopes()` is rerun?

Given the current documentation, I think the least surprising behavior would be to preserve the existing target state. Skipping or explicitly rejecting the collision would also be reasonable, but silently reactivating the target seems inconsistent with the documented claim that consolidation does not reactivate non-active networks and that rerunning the operation is safe.

With that case resolved and tested, I would be comfortable approving the PR.

